PR #46, opened 2016-12-04
run init_roots() before flask

busla:


vthorsteinsson:
Not necessary after the latest updates where scraperinit.py is run via ENTRYPOINT in the Docker container greynir/web.

PR #64, opened 2018-11-05
Fixed broken image problem, added image blacklisting + cached image URLs expire

sveinbjornt:
This should fix the broken image issue. The logic is the following:

If the image fails to load client-side, it is immediately hidden and we're notified via /brokenimage, which receives URL and name.

If name and URL exists in Links table, and hasn't been updated very recently, we send a HEAD request to the URL to verify that it is broken. If it is, we purge the entry and make a new query to Google.

The new image URL is then returned to the client, and the attribute of the image updated. The image is then shown again.


vthorsteinsson:
Looks really good! Only question is whether line 101 in `getimage.py` should use `logging.warning()` instead of `print()`.

sveinbjornt:
Added blacklisting of images via "Röng mynd?" overlaid onhover link for images. Ready for merging.

sveinbjornt:
Added blacklisting of images via "Röng mynd?" overlaid onhover link for images. Ready for merging after review.

getimage.py
```diff
@@ -1,149 +1,267 @@
...
+
+        if from_cache:
+            q = session.query(Link.content, Link.timestamp) \
+                .filter(Link.ctype == ctype) \
+                .filter(Link.key == name) \
+                .one_or_none()
+            if q is not None:
+                # Found in cache
+                if datetime.utcnow() - q.timestamp > timedelta(days=_CACHE_EXPIRATION_DAYS):
+                    _purge_single(name, ctype=ctype, session=session)
```
vthorsteinsson: Shouldn't this be:

```suggestion
                    _purge_single(name, ctype=ctype, enclosing_session=session)
```

PR #66, opened 2018-11-12
IPython Shell

sveinbjornt:
Run ./shell.sh from repo root to launch an IPython shell, vastly superior to the standard Python REPL. Features include syntax highlighting, auto-pretty-printing, auto-indentation, smart autocompletion, smart persistent history across sessions, easy timing of functions, integrated access to pdb and the profiler, various introspection tools etc.

Auto-reloading of modules prior to every command can be enabled by typing '%autoreload 2'

I've configured the shell to automatically import the database models and create a (commit-disabled) database session when launched. We can set it up to auto-import anything we want by mucking around with the .ipython.py config file. Local user settings can be configured in ~/.ipython/profile_default.

See https://ipython.readthedocs.io/en/stable/ for details.

### DIFF
--- a/.ipython.py
+++ b/.ipython.py
@@ -1,0 +1,14 @@
+# Reynir configuration file for ipython.
+
+from platform import python_version, python_implementation
+
+c.TerminalInteractiveShell.confirm_exit = False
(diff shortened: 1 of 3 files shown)


vthorsteinsson:
Very cool!

PR #68, opened 2018-11-15
Query autosuggestions

sveinbjornt:
Adds query autocomplete suggestions for questions starting with "Hver er " (persons/titles) and "Hvað er" (entities). Person/title/entity suggestions are ordered by related article count, which seems to deliver quite sensible results.

Uses the jQuery autocomplete plugin. Implemented aggressive caching both client- and server-side. Server-side caching was added using the Flask-Caching module (which is now included in requirements.txt). I'm still not up to speed on the caching setup for the nginx, Gunicorn, Flask pipeline, so by all means let me know if there's a better way to do this.


main.py
```diff
@@ -829,6 +833,64 @@ def reportimage():
...
+    """ Return suggestions for query field autocompletion """
+    limit = request.args.get('limit', limit)
+    txt = request.args.get("q", "").strip()
+
+    suggestions = list()
+    whois_prefix = "Hver er "
+    whatis_prefix = "Hvað er "
+
+    prefix = None
+    if txt.startswith(whois_prefix):
```
vthorsteinsson: Probably should be `txt.lower().startswith(whois_prefix)` - OK to be permissive on the upper/lower case

main.py
```diff
@@ -829,6 +833,64 @@ def reportimage():
...
 
+@app.route("/suggest", methods=["GET"])
+@cache.cached(timeout=300, key_prefix='suggestions', query_string=True)
+def suggest(limit=10):
+    """ Return suggestions for query field autocompletion """
+    limit = request.args.get('limit', limit)
+    txt = request.args.get("q", "").strip()
+
+    suggestions = list()
+    whois_prefix = "Hver er "
```
vthorsteinsson: Probably should be "hver er " (lower case) - see next comment

main.py
```diff
@@ -829,6 +833,64 @@ def reportimage():
...
+    whatis_prefix = "Hvað er "
```
vthorsteinsson: "hvað er " - see next comment

main.py
```diff
@@ -829,6 +833,64 @@ def reportimage():
...
+
+        # Hver er Jón Jónsson ?
+        if prefix is whois_prefix and name[0].isupper():
+            # Find matching persons ordered by number of 
+            # related articles over the past year
+            q = (session.query(Person.name, dbfunc.count(Article.id).label('total'))
+                .filter(Person.name.ilike(name + '%'))
+                .join(Article)
+                .filter(Article.timestamp > datetime.utcnow() - timedelta(days=365))
+                .group_by(Person.name)
```
vthorsteinsson: Don't you need to group on a sum of total?

main.py
```diff
@@ -829,6 +833,64 @@ def reportimage():
...
+                .order_by(desc('total'))
+            )
+        # Hver er seðlabankastjóri?
+        elif prefix is whois_prefix:
+            # Find matching title
+            q = (session.query(Person.title, dbfunc.count(Article.id).label('total'))
+                .filter(Person.title.ilike(name + '%'))
+                .join(Article)
+                .group_by(Person.title)
```
vthorsteinsson: Again here: GROUP BY title, SUM(total) AS total_cnt

static/js/main.js
```diff
@@ -534,6 +536,46 @@ function getUrlVars() {
...
+      return;
+   }
+   // Cancel any active request
+   if (autoCompleteLookup.req) {
+      autoCompleteLookup.req.abort();
+   }
+   // Ajax request to server
+   autoCompleteLookup.req = $.ajax({
+      type: 'GET',
+      url: "/suggest?q="+ query,
```
vthorsteinsson: Don't we need to urlencode or otherwise encode query?

static/js/main.js
```diff
@@ -534,6 +536,46 @@ function getUrlVars() {
...
+      dataType: "json",
+      success: function(json) {
+         autoCompleteLookup.cache[query] = json;
+         done(json);
+      },
+      error: function(ajaxContext) {
+         console.log(ajaxContext.responseText)
```
vthorsteinsson: What does the UI show in the case of an error here?

sveinbjornt: Nothing. Autocompletion of queries is a nice-to-have, non-essential feature. If the request fails, the suggestions simply don't appear. To my mind, it makes no sense to notify the user client-side.

main.py
sveinbjornt: Maybe I'm just dim-witted today, but I don't see the problem with the query. The SQL generated by SQLAlchemy is "SELECT persons.name AS persons_name, count(articles.id) AS total FROM persons JOIN articles ON articles.url = persons.article_url GROUP BY persons.name ORDER BY total DESC;"

Isn't that exactly what we want?

PR #69, opened 2018-11-16
Image for persons in onhover box

sveinbjornt:
Fetch thumbnail image for persons in word onhover box. For now, I've set it to only show an image if an image URL exists in the cache (i.e. no Google API calls are triggered on hover). We can change this later if we want, but let's first see how this turns out.

Also, image search now prioritises results on Icelandic-language pages and servers with an Icelandic geolocation.

### DIFF
--- a/images.py
+++ b/images.py
@@ -110,6 +110,9 @@ def get_image_url(name, size="large", thumb=False, enclosing_session=None, from_
                 num = _NUM_IMG_URLS,
                 start = 1,
                 imgSize = size,
+                imgType = "face",   # Only images with faces
+                lr = "lang_is",     # Higher priority for Icelandic language pages
+                gl = "is",          # Higher priority for .is results
                 searchType = "image",
                 cx = _CX,
                 key = _API_KEY
--- a/images.py
+++ b/images.py
@@ -69,23 +69,26 @@ _NUM_IMG_URLS = 10
 Img = namedtuple('Img', ['src', 'width', 'height', 'link', 'origin', 'name'])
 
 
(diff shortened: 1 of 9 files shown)


images.py
```diff
@@ -69,24 +69,26 @@ def _server_query(url, q):
 Img = namedtuple('Img', ['src', 'width', 'height', 'link', 'origin', 'name'])
 
 
-def get_image_url(name, size="large", enclosing_session=None, from_cache=True):
+def get_image_url(name, hints=[], size="large", thumb=False,
+                  enclosing_session=None, cache_only=False):
```
vthorsteinsson: Use black for the indent style

images.py
```diff
@@ -69,24 +69,26 @@ def _server_query(url, q):
...
-                .filter(Link.ctype == ctype) \
-                .filter(Link.key == name) \
-                .one_or_none()
-            if q is not None:
-                # Found in cache
-                if datetime.utcnow() - q.timestamp > timedelta(days=_CACHE_EXPIRATION_DAYS):
-                    _purge_single(name, ctype=ctype, enclosing_session=session)
-                else:
-                    jdoc = q.content
+        q = session.query(Link.content, Link.timestamp) \
```
vthorsteinsson: Black will also fix these line continuations

templates/hover-infobox.html
```diff
@@ -0,0 +1,23 @@
+
+<!-- Popup with info about the word being hovered over -->
+<div id="info">
+	<table>
```
vthorsteinsson: Indent with 3 spaces for consistency

PR #71, opened 2018-11-27
Pagination

sveinbjornt:
Adds pagination to Nýjast. Also fixes some layout stuff, adds localized date strings, padding to tables and links to source article.

I removed the deduplication logic in top_news() since it made rational pagination impossible. We should be handling removal of duplicates in the database anyway, as we discussed.


main.py
```diff
@@ -284,40 +291,35 @@ def time(self):
...
-            d = ArticleDisplay(
-                heading=a.heading,
-                timestamp=a.timestamp,
-                url=a.url,
-                uuid=a.id,
-                num_sentences=a.num_sentences,
-                num_parsed=a.num_parsed,
-                icon=icon,
-            )
+                if datetime.today().year is self.timestamp.year:
```
vthorsteinsson: Should this be `==` instead of `is`?

main.py
```diff
@@ -1039,32 +1041,33 @@ def apidoc():
...
-    articles = top_news(topic=topic, start=start)
-    now = datetime.utcnow()
+    try:
+        offset = max(0, int(request.args.get("offset", 0)))
+        limit = max(0, int(request.args.get("limit", _TOP_NEWS_LENGTH)))
+    except:
+        offset = 0
+        limit = _TOP_NEWS_LENGTH
+
+    articles = top_news(topic=topic, offset=offset)
```
vthorsteinsson: Clamp `limit` to a maximum value here (avoiding cases where people manually put in 100 billion or whatever)?

vthorsteinsson: ...and the same applies to `offset` also, btw.

sveinbjornt: Well spotted wrt limit. Offset could be 100 billion without trouble, would just return 0 results.

PR #73, opened 2018-12-05
Location processing

sveinbjornt:


geo.py
```diff
@@ -0,0 +1,161 @@
...
+def coords_for_street_name(street_name, placename=None, placename_hints=[]):
+    """ Return coordinates for an Icelandic street name as a tuple """
+
+    addresses = iceaddr_lookup(street_name, placename=placename, limit=100)
+    places = set(a["stadur_nf"] for a in addresses if a.get("stadur_nf"))
+    addr = None
+
+    # Street name only exists in one place
+    if len(places) == 1:
+        addr = addresses[0]
```
vthorsteinsson: What about the case where `addresses[0]["stadur_nf"]` is `None` or empty? From the set generator expression it looks like that could be the case.

geo.py
```diff
@@ -0,0 +1,161 @@
...
+        return (addr["lat_wgs84"], addr["long_wgs84"])
+
+    return None
+
+
+def country_name_for_isocode(iso_code, lang=ICELANDIC_LANG_ISOCODE):
+    """ Return country name for an ISO 3166-1 alpha-2 code """
+    assert len(iso_code) == 2 and iso_code.isupper()
+    assert len(lang) == 2 and lang.islower()
+    assert lang in available_languages()
```
vthorsteinsson: Consider returning `None` in at least some of these error cases?

geo.py
```diff
@@ -0,0 +1,161 @@
...
+
+def isocode_for_country_name(country_name, lang=ICELANDIC_LANG_ISOCODE):
+    """ Return ISO 3166-1 alpha-2 code for a country 
+        name in the specified language (two-char ISO 639-1) """
+    assert len(lang) == 2 and lang.islower()
+    assert lang in available_languages()
+
+    countries = countries_for_language(lang)
+    for iso_code, name in countries:
+        if name == country_name:
```
vthorsteinsson: Case is significant here. Perhaps state that in the function docstring?

geo.py
```diff
@@ -0,0 +1,161 @@
...
+            "Makaó": "MO",
+            "England": "GB",
+            "Skotland": "GB",
+            "Bosnía": "BA",
+            "Hersegóvína": "BA",
+            "Palestína": "PS",
+        }
+    }
+
+    if additions.get(lang):
```
vthorsteinsson: `if lang in additions` is clearer

geo.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def lookup(pn):
+        a = iceaddr_lookup(
+            addr["street"],
+            number=addr.get("number"),
+            letter=addr.get("letter"),
+            placename=pn,
+            limit=100,
+        )
+        return a[0] if len(a) == 1 else None
```
vthorsteinsson: Should this be `len(a) >= 1`?

geo.py
```diff
@@ -0,0 +1,161 @@
...
+
+    # Look up with the (optional) placename provided
+    res = lookup(placename)
+
+    # If no single address found, try to disambiguate using placename hints
+    if not res:
```
vthorsteinsson: `if not res and placename_hints:`

geo.py
```diff
@@ -0,0 +1,161 @@
...
+
+    comp = addrstr.split(" ")
+    if len(comp) == 1:
+        return addr
+
+    last = comp[-1]
+    r = re.search(r"^(\d+)([a-zA-Z]?)$", last)
+    if r:
+        addr["number"] = int(r.group(1))
+        addr["letter"] = r.group(2) if r.group(2) else None
```
vthorsteinsson: Could also say `= r.group(2) or None`

geo.py
```diff
@@ -0,0 +1,161 @@
...
+                break
+
+    return res
+
+
+def parse_address_string(addrstr):
+    """ Break Icelandic address string down to its components """
+    addr = {"street": addrstr}
+
+    comp = addrstr.split(" ")
```
vthorsteinsson: Why not just `.split()`? The semantics of `.split(" ")` are slightly different.

main.py
```diff
@@ -379,6 +380,53 @@ def is_better_title(new_title, old_title):
...
+            .order_by(desc(Article.timestamp))
+        )
+
+    if kind:
+        q = q.filter(Location.kind == kind)
+
+    toplist = []
+    for l in q.limit(limit).all():
+        zoom = "6z" if l.kind == "country" else "16z"
+        mapurl = "http://google.com/maps/@{0},{1},{2}".format(l[4], l[5], zoom)
```
vthorsteinsson: Does `l.latitude, l.longitude` not work? And should it not be `https://` rather than `http://`?

main.py
```diff
@@ -971,6 +1019,15 @@ def suggest(limit=10):
     return better_jsonify(suggestions=suggestions)
 
 
+@app.route("/locations", methods=["GET"])
+@max_age(seconds=5 * 60)
+def locations():
+    kind = request.args.get("kind")
+    locs = top_locations(limit=100, kind=kind)
```
vthorsteinsson: Going forward, 100 locations is probably too much - include as a constant at the top of the source file?

processor.py
```diff
@@ -193,14 +219,23 @@ def iter_parsed_articles():
                 self.go_single(url)
         else:
             # Use a multiprocessing pool to process the articles
-            pool = Pool() # Defaults to using as many processes as there are CPUs
+            pool = Pool(
+                self.workers
+            )  # Defaults to using as many processes as there are CPUs
```
vthorsteinsson: Fix formatting here (Black has a bug with inline comments such as this)

processors/locations.py
```diff
@@ -0,0 +1,169 @@
...
+        return
+
+    print(locs)
+
+    # Find all placenames mentioned in article
+    # We can use them to disambiguate addresses and street names
+    placenames = [p.name for p in locs if p.kind == "placename"]
+
+    # Get info about each location and save to database
+    for loctuple in locs:
```
vthorsteinsson: `for name, kind in locs:` ?

processors/locations.py
```diff
@@ -0,0 +1,169 @@
...
+        loc = loctuple._asdict()
+        kind = loc["kind"]
+        coords = None
+
+        # Heimilisfang
+        if kind == "address":
+            loc["country"] = ICELAND_ISOCODE
+            info = icelandic_addr_info(loc["name"], placename_hints=placenames)
+            if info and info.get("lat_wgs84") and info.get("long_wgs84"):
```
vthorsteinsson: Create a utility function in `geo` to extract lat and long from dict?

processors/locations.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # Götuheiti
+        elif kind == "street":
+            # All the street names in BÍN are Icelandic
+            loc["country"] = ICELAND_ISOCODE
+            coords = coords_for_street_name(loc["name"], placename_hints=placenames)
+
+        # Örnefni
+        elif kind == "placename":
+            if loc["name"] in PLACENAME_BLACKLIST:
```
vthorsteinsson: Is upper and lower case definitely not significant here?

processors/locations.py
```diff
@@ -0,0 +1,169 @@
...
+                continue
+
+        if coords:
+            (loc["latitude"], loc["longitude"]) = coords
+
+        loc["article_url"] = url
+        loc["timestamp"] = datetime.utcnow()
+
+        l = Location(**loc)
```
vthorsteinsson: Cool

processors/locations.py
```diff
@@ -0,0 +1,169 @@
...
+from scraperdb import Location
+from geo import (
+    coords_for_country,
+    coords_for_street_name,
+    icelandic_addr_info,
+    isocode_for_country_name,
+    ICELAND_ISOCODE,
+)
+
+PLACENAME_BLACKLIST = ["Staður", "Eyjan", "Fjöll", "Bæir"]
```
vthorsteinsson: `frozenset(...)`

processors/locations.py
```diff
@@ -0,0 +1,169 @@
...
+    l = Loc(name=addr_nom, kind="address")
+    result._state["locations"].add(l)
+
+
+def _process(node, params, result):
+    """ Look up meaning in BÍN, add as location if in right category """
+    state = result._state
+
+    loc_fl = ["göt", "lönd", "örn"]
+    binfl2kind = dict(zip(loc_fl, ["street", "country", "placename"]))
```
vthorsteinsson: Move this to constants at the top of the file or just before `def _process(...):`

processors/locations.py
sveinbjornt: Well, at this point we're working with the canonical nominative form of the placename (in fl "örn) from BÍN, which should be capitalised.

sveinbjornt: But point taken, no harm in having the comparison case insensitive.

geo.py
sveinbjornt: Well spotted!

PR #74, opened 2018-12-17
Locations processing + first iteration of /locations interface

sveinbjornt:
Location processing + new /locations page showing recent locations in news.


geo.py
```diff
@@ -0,0 +1,247 @@
...
+            if len(places) == 1:
+                addr = addresses[0]
+                break
+
+    return coords_from_addr_info(addr)
+
+
+def coords_from_addr_info(info):
+    """ Get coordinates from the address dict provided by iceaddr """
+    if info and info.get("lat_wgs84") and info.get("long_wgs84"):
```
vthorsteinsson: Better: `if info is not None and "lat_wgs84" in info and "long_wgs84" in info:`

geo.py
```diff
@@ -0,0 +1,247 @@
...
+
+def country_name_for_isocode(iso_code, lang=ICELANDIC_LANG_ISOCODE):
+    """ Return country name for an ISO 3166-1 alpha-2 code """
+    assert len(iso_code) == 2
+    assert len(lang) == 2
+
+    iso_code = iso_code.upper()
+    lang = lang.lower()
+
+    if not lang in available_languages():
```
vthorsteinsson: Better: `if lang not in available_languages():`

geo.py
```diff
@@ -0,0 +1,247 @@
...
+    return countries.get(iso_code)
+
+
+def isocode_for_country_name(country_name, lang=ICELANDIC_LANG_ISOCODE):
+    """ Return the ISO 3166-1 alpha-2 code for a country 
+        name in the specified language (two-char ISO 639-1) """
+    assert len(lang) == 2
+
+    lang = lang.lower()
+    if not lang in available_languages():
```
vthorsteinsson: Better: `if lang not in available_languages():`

geo.py
```diff
@@ -0,0 +1,247 @@
...
+        return None
+
+    countries = countries_for_language(lang)
```
vthorsteinsson: Consider making `countries_for_language()` return a `dict`

geo.py
```diff
@@ -0,0 +1,247 @@
...
+    for iso_code, name in countries:
+        if name == country_name:
+            return iso_code
+
+    additions = {
```
vthorsteinsson: Put this constant at module scope, at top of file?

vthorsteinsson: ...or, even better going forward, in a `.conf` file that is read in `settings.py`.

geo.py
```diff
@@ -0,0 +1,247 @@
...
+
+def parse_address_string(addrstr):
+    """ Break Icelandic address string down to its components """
+    addr = {"street": addrstr}
+
+    comp = addrstr.split()
+    if len(comp) == 1:
+        return addr
+
+    last = comp[-1]
```
vthorsteinsson: Add comment here, nonobvious code

images.py
```diff
@@ -55,16 +58,38 @@ def _server_query(url, q):
     return doc
 
 
-# The Custom Search identifier
-_CX = "001858240983628375092:9aogptqla5e"
 # The Google API identifier (you must obtain your own key if you want to use this code)
 _API_KEY = ""
+_API_KEY_PATH = "resources/GoogleServerKey.txt"
```
vthorsteinsson: Use `os.path.join()` here (don't forget Windows! :-) )

images.py
```diff
@@ -55,16 +58,38 @@ def _server_query(url, q):
...
+
+
+def _get_API_key():
+    """ Read Google API key from file """
+    global _API_KEY
+    if _API_KEY:
```
vthorsteinsson: Better: `if not _API_KEY: try [...]`, then end with `return _API_KEY` in all cases

images.py
vthorsteinsson: ... and actually you probably need:
```
_PATH = os.path.dirname(__file__) or "."
_API_KEY_PATH = os.path.join(_PATH, "resources", "GoogleServerKey)
```

images.py
```diff
@@ -55,16 +58,38 @@ def _server_query(url, q):
...
+        # You need to obtain your own key and put it in
+        # _API_KEY_PATH if you want to use this code
+        with open(_API_KEY_PATH) as f:
+            _API_KEY = f.read().rstrip()
+    except FileNotFoundError as ex:
+        _API_KEY = ""
+    return _API_KEY
+
+
+# Custom Search identifier
```
vthorsteinsson: Comment this - what is this magic constant?

images.py
```diff
@@ -85,9 +110,11 @@ def get_image_url(
         )
         if q is not None:
             # Found in cache. If the result is old, purge it
-            expired = datetime.utcnow() - q.timestamp > timedelta(days=_CACHE_EXPIRATION_DAYS)
+            expired = datetime.utcnow() - q.timestamp > timedelta(
+                days=_CACHE_EXPIRATION_DAYS
```
vthorsteinsson: Funny indent

images.py
```diff
@@ -251,6 +269,36 @@ def _purge():
...
+
+
+def get_staticmap_image(latitude, longitude, zoom=6, width=180, height=180):
+    """ Request image from Google Static Maps API, return image data as bytes """
+    key = _get_API_key()
+    if not key:
+        return None
+
+    url = STATICMAP_URL.format(zoom, width, height, key, latitude, longitude)
+    # TODO: Use urllib instead of requests here
```
vthorsteinsson: I'm not sure how Gunicorn/gevent/eventlet handle this, i.e. whether a call to `requests` blocks the web server worker. I suspect it does.

images.py
```diff
@@ -251,6 +269,36 @@ def _purge():
...
+    try:
+        r = requests.get(url, stream=True)
+    except:
+        logging.warning(str(e))
```
vthorsteinsson: Missing `except Exception as e:`

main.py
```diff
@@ -213,12 +221,16 @@ def better_jsonify(**kwargs):
...
+# Default number of top news items to show in /news
 _TOP_NEWS_LENGTH = 20
 
-# Default number of top persons to show in front page list
+# Default number of top persons to show in /people
 _TOP_PERSONS_LENGTH = 20
 
+# Default number of top locations to show in /locations
+_TOP_LOCATIONS_LENGTH = 20
+_TOP_LOCATIONS_PERIOD = 1
```
vthorsteinsson: Which units?

main.py
```diff
@@ -971,6 +1053,72 @@ def suggest(limit=10):
     return better_jsonify(suggestions=suggestions)
 
 
+@app.route("/locations", methods=["GET"])
+@max_age(seconds=60 * 5)
+def locations():
+    """ Render locations page """
+    kind = request.args.get("kind")
+    period = request.args.get("period")
+
```
vthorsteinsson: Probably need more error checking on `kind` and `period` arguments here, since they are effectively user-supplied

main.py
```diff
@@ -971,6 +1053,72 @@ def suggest(limit=10):
...
+    days = _TOP_LOCATIONS_PERIOD
+    if period == "week":
+        days = 7
+
+    kind = kind if kind in LOCATION_TAXONOMY else None
+
+    with SessionContext(commit=False) as session:
+        locs = top_locations(enclosing_session=session, kind=kind, days=days)
+
+    return render_template("locations/locations.html", locations=locs, period=period)
```
vthorsteinsson: Is a path with forward slashes definitely allowed under Windows?

main.py
```diff
@@ -971,6 +1053,72 @@ def suggest(limit=10):
...
+@app.route("/staticmap", methods=["GET"])
+@cache.cached(timeout=60 * 60 * 24, key_prefix="staticmap", query_string=True)
+def staticmap():
+    """ Proxy for Google Static Maps API """
+    try:
+        lat = float(request.args.get("lat"))
+        lon = float(request.args.get("lon"))
+        zoom = int(request.args.get("z"))
+    except:
+        return server_error(500)
```
vthorsteinsson: Perhaps a 500 error is not correct here - this is really about a wrong or malformed URL

main.py
```diff
@@ -971,6 +1053,72 @@ def suggest(limit=10):
...
+
+    return render_template("locations/locations.html", locations=locs, period=period)
+
+
+@app.route("/staticmap", methods=["GET"])
+@cache.cached(timeout=60 * 60 * 24, key_prefix="staticmap", query_string=True)
+def staticmap():
+    """ Proxy for Google Static Maps API """
+    try:
+        lat = float(request.args.get("lat"))
```
vthorsteinsson: Include reasonable defaults in the `.get()` calls?

main.py
```diff
@@ -971,6 +1053,72 @@ def suggest(limit=10):
...
+    if name == "Ísland":
+        kind = "country"
+
+    if name and kind and kind in LOCATION_TAXONOMY:
+        loc = location_info(name, kind)
+        resp["found"] = True
+        resp["country"] = loc.get("country")
+        lat, lon = loc.get("latitude"), loc.get("longitude")
+        if lat and lon:
+            zoom4kind = {"street": 12, "placename": 5, "country": 2}
```
vthorsteinsson: Include as a constant at module level

main.py
```diff
@@ -971,6 +1053,72 @@ def suggest(limit=10):
...
+STATIC_MAP_URL = "/staticmap?lat={0}&lon={1}&z={2}"
+
+
+@app.route("/locinfo", methods=["GET"])
+@cache.cached(timeout=60 * 60, key_prefix="locinfo", query_string=True)
+def locinfo():
+    """ Return info about a location as JSON """
+    resp = dict(found=False)
+
+    name = request.args.get("name")
```
vthorsteinsson: Error-check arguments

static/js/page.js
```diff
@@ -102,6 +116,11 @@ function queryEntity(name) {
    window.location.href = "/?f=q&q=" + encodeURIComponent("Hvað er " + name + "?");
 }
 
+function queryLocation(name) {
+   // Implement me!
```
vthorsteinsson: Add `TODO:` in comment

images.py
sveinbjornt: How about using pathlib? Do we need to support Python 3.3 and earlier?

main.py
sveinbjornt: Yes, they are apparently processed by Jinja and not treated naively as filesystem paths.

images.py
sveinbjornt: Yeah, it blocks the worker. Do you have any suggestions on how to handle this differently? There's a variety of async requests-like libraries out there.

images.py
vthorsteinsson: Nope, Python >= 3.4 is the requirement

images.py
vthorsteinsson: Need to investigate whether using urllib would be compatible with the monkeypatching scheme of greenlets/eventlets.

vthorsteinsson: ...but the futureproof solution is to migrate to an async based web server framework, such as `responder`.

PR #91, opened 2019-08-21
Added arithmetic and time query modules, with tests.

sveinbjornt:
* New query module to handle arithmetic queries
* Improved time query module (formerly clock.py)


queries/arithmetic.py
```diff
@@ -0,0 +1,410 @@
...
+            r = calc_arithmetic(q, result)
+            if r is not None:
+                q.set_answer(*r)
+                q.set_key(result.get("qkey"))
+            else:
+                raise Exception("Arithmetic calculation failed")
+        except AssertionError:
+            raise
+        except Exception as e:
+            raise
```
vthorsteinsson: Need to choose whether to re-raise or to do `q.set_error()` (the latter is probably correct for release)

PR #97, opened 2019-09-17
Distance module + more

sveinbjornt:
* New distance query module (not perfect, overly reliant on Google's API delivering sensible results, but it's a start)
* New geography query module (very basic, just capital cities but more to come!)
* Location support in weather module ("Hvernig er veðrið á Ísafirði?")
* Updated cityloc package dependency
* Code to talk to Google's geocode API now in queries/__init__.py so it can be used by all modules if needed
* Added TTL caching to get_synthesized_text_url function in speech.py (at this point unclear what the implications of higher speech URL expiry time on AWS servers means, will investigate)


queries/distance.py
```diff
@@ -0,0 +1,134 @@
...
+def handle_plain_text(q):
+    """ Handle a plain text query, contained in the q parameter """
+    ql = q.query_lower.rstrip("?")
+
+    remote_loc = None
+    for rx in _QREGEXES:
+        m = re.search(rx, ql)
+        if m:
+            remote_loc = m.group(1)
+            break
```
vthorsteinsson: Could do `for...else` here:

```
for rx in _QREGEXES:
    m = re.search(rx, ql)
        if m:
            remote_loc = m.group(1)
            break
else:
    return False
```

queries/distance.py
```diff
@@ -0,0 +1,134 @@
...
+
+    if not remote_loc:
+        return False
+
+    try:
+        answ = answer_for_remote_loc(remote_loc, q)
+    except Exception as e:
+        logging.warning("Exception looking up addr in geocode API: {0}".format(e))
+        answ = None
```
vthorsteinsson: Maybe set `q.error` here?

queries/geography.py
```diff
@@ -0,0 +1,79 @@
...
+from reynir.bindb import BIN_Db
+
+
+_GEO_QTYPE = "Geography"
+
+
+def handle_plain_text(q):
+    """ Handle a plain text query, contained in the q parameter """
+    ql = q.query_lower.rstrip("?").strip()
+    pfx = "hver er höfuðborg "
```
vthorsteinsson: Add support for 'Hver er höfuðborgin í X'?

queries/geography.py
```diff
@@ -0,0 +1,79 @@
...
+        country = q.query[len(pfx) :].strip()
+        if not len(country):
+            return False
+
+        country = country[0].upper() + country[1:]  # Capitalize first char
+        # TODO: This only works for single-word country names, fix that
+        # Transform country name from genitive to nominative
+        bres = BIN_Db().lookup_nominative(country, cat="no")
+        words = [m.stofn for m in bres]
+        if not words:
```
vthorsteinsson: It would be safer to put one line earlier: `if not bres: return False`

queries/geography.py
```diff
@@ -0,0 +1,79 @@
...
+        if not cc:
+            return False
+
+        # Find capital city, given the country code
+        capital = capital_for_cc(cc)
+        if not capital:
+            return False
+
+        # Use the Icelandic name for the city
+        ice_cname = icelandic_city_name(capital["name_ascii"])
```
vthorsteinsson: Funny key name: `name_ascii`!

queries/weather.py
```diff
@@ -102,6 +104,12 @@
 QWeatherCapitalRegion →
     "á" "höfuðborgarsvæðinu" | "í" "reykjavík"
 
+QWeatherAnyLoc →
+    QWeatherCountry | QWeatherCapitalRegion | QWeatherOpenLoc
```
vthorsteinsson: These alternatives look like they might be ambiguous with respect to each other. Might be good to have `>` signs between alternatives in the production, rather than `|`, to indicate absolute priority. Or, otherwise, to give them different scores with the `$score()` pragma.

queries/weather.py
```diff
@@ -176,16 +184,40 @@ def _round_to_nearest_hour(t):
...
+        to the location associated with the query (i.e. either user location 
+        coordinates or a specific placename) """
     loc = query.location
 
+    # User asked about a specific location
+    # Try to find a matching Icelandic placename
+    if "location" in result and result.location != "Ísland":
+
+        # Some strings should never be interpreted as Icelandic placenames
+        if result.location in ICE_PLACENAME_BLACKLIST:
```
vthorsteinsson: Is this required here?

queries/weather.py
```diff
@@ -176,16 +184,40 @@ def _round_to_nearest_hour(t):
...
+            return None
+
+        # Unfortunately, many foreign country names are also Icelandic
+        # placenames, so we automatically exclude country names.
+        cc = isocode_for_country_name(result.location)
```
vthorsteinsson: Is this required here?

speech.py
```diff
@@ -74,7 +78,33 @@ def _intialize_client():
     return boto3.Session(**aws_config).client("polly")
 
 
-# TODO: Add caching for identical responses
+def ttlcache(ttl=datetime.timedelta(seconds=600)):
```
vthorsteinsson: Looks good except that I see no code anywhere to cull expired data from the cache?

queries/weather.py
sveinbjornt: As long as the weather module can't provide information for locations outside Iceland, we don't want queries about foreign places (e.g. "á Spáni", "í París") to return data for Icelandic placenames with the same name.

vthorsteinsson: Understood!

speech.py
sveinbjornt: Yeah, haha, in any case silly to roll our own. Replacing with TTLCache from the cachetools package.

PR #99, opened 2019-10-02
Clear query history + query chart

sveinbjornt:


routes/api.py
```diff
@@ -379,3 +379,29 @@ def query_api(version=1):
...
+    client_id = request.values.get("client_id")
+
+    if action != "clear" or not client_type or not client_id or len(client_id) < 18:
+        return better_jsonify(valid=False, reason="Missing parameters")
+
+    with SessionContext(commit=True) as session:
+        session.execute(
+            Query.table()
+            .delete()
+            .where(Query.client_type == client_type and Query.client_id == unique_id)
```
vthorsteinsson: Should probably be `Query.client_id == client_id`

routes/stats.py
```diff
@@ -92,6 +101,10 @@ def chart_stats(session=None, num_days=7):
             percent = round((parsed / sent) * 100, 2) if sent else 0
             parsed_data.append(percent)
 
+            # Get query count for day
```
vthorsteinsson: A bit ugly, since this executes a separate SQL query for each day; but maybe OK for this non-critical use case.

PR #100, opened 2019-10-04
Great big whale of a pull request

sveinbjornt:
* New telephone module
* New wikipedia module
* New random module to "roll a die" or provide a number in a given range
* New opinion module through which Embla expresses no strong feelings one way or the other on a wide variety of subjects :)
* Many, many additions to the special query module
* Query handlers can now set an URL associated with the query, which is then delivered to client
* Geography module now uses parse trees and handles a broader range of queries
* Fixed some issues in weather module + upgraded iceweather pkg dependency
* Various minor tweaks and improvements in most of the other query modules
* Additions to Vocab.conf


queries/time.py
```diff
@@ -39,6 +39,18 @@
 _TIME_QTYPE = "Time"
 
 
+_TIME_QUERIES = (
```
vthorsteinsson: Here, a `frozenset()` would be better

queries/weather.py
```diff
@@ -104,7 +105,8 @@
     "á" "landinu" | "á" "íslandi" | "hér" "á" "landi" | "á" "landsvísu"
 
 QWeatherCapitalRegion →
-    "á" "höfuðborgarsvæðinu" | "í" "reykjavík"
+    "á" "höfuðborgarsvæðinu" | "fyrir" "höfuðborgarsvæðið" 
+    | "í" "reykjavík" | "fyrir" "reykjavík"
```
vthorsteinsson: Add `| "í" "höfuðborginni" | "fyrir" "höfuðborgina"`

queries/wiki.py
```diff
@@ -0,0 +1,156 @@
+"""
+
+    Reynir: Natural language processing for Icelandic
+
+    Arithmetic query response module
```
vthorsteinsson: Update module description

queries/wiki.py
```diff
@@ -0,0 +1,156 @@
...
+)
+
+
+def QWikiQuery(node, params, result):
+    # Set the query type
+    result.qtype = _WIKI_QTYPE
+    result.qkey = result["subject_nom"]
+
+
+def Nl(node, params, result):
```
vthorsteinsson: Might be better do define the function `QWikiSubject()` to handle this, since it may become more complicated going forward

queries/wiki.py
```diff
@@ -0,0 +1,156 @@
...
+
+def _query_wiki_api(subject):
+    url = _WIKI_API_URL.format(subject)
+    return query_json_api(url)
+
+
+def get_wiki_summary(subject_nom, subject_dat):
+    res = _query_wiki_api(subject_nom)
+
+    not_found = "Ekkert fannst um {0} í Wikipedíu".format(subject_dat)
```
vthorsteinsson: Better: `"Ég fann ekkert um {0} í Wikipedíu"`

queries/geography.py
```diff
@@ -23,58 +23,100 @@
...
+
+def QGeoContinentQuery(node, params, result):
+    result["geo_qtype"] = "continent"
+
+
+def QGeoSubject(node, params, result):
+    n = result._nominative
+    if n:
+        n = n.replace(" - ", "-")
+        n = n[:1].capitalize() + n[1:]
```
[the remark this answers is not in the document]
sveinbjornt: the capitalize() function not only capitalizes the first character but also (annoyingly) makes the rest of the string lowercase. But n[0] is better than n[:1] :)

queries/special.py
```diff
@@ -144,56 +212,284 @@ def _rudeness():
...
     "ertu með kímnigáfu": {
-        "answer": "Afar takmarkaða.",
-        "voice": "Já, en afar takmarkaða",
+        "answer": "Já, en afar takmarkaða.",
+    },
+    "er þú með kímnigáfu": {
+        "answer": "Já, en afar takmarkaða.",
+    },
+    "ertu með húmor": {
+        "answer": "Já, en afar takmarkaðann.",
```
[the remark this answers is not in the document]
sveinbjornt: Jeminn, ugh! **skammast sín**

PR #106, opened 2019-11-21
Qimprov

sveinbjornt:
Strip certain prefixes from query strings prior to processing, e.g. "hæ embla", "embla"


query.py
```diff
@@ -211,6 +215,20 @@ def __init__(self, session, query, voice, auto_uppercase, location, client_id):
         # This should be a dict that can be represented in JSON
         self._context = None
 
+
+    def _preprocess_query_string(self, q):
+        """ Preprocess the query string prior to further analysis. """
+        if not q:
+            return q
+        pfxrx = "|".join(IGNORED_QUERY_PREFIXES)
```
vthorsteinsson: Better would be `"^(hæ embla|embla|hey embla)\s*"`, this ensures that we only replace at the beginning of the string and that any following whitespace is deleted too (makes the `.strip()` unnecessary)

sveinbjornt: Whoops, missing ^, clumsy of me. Fixed.

PR #122, opened 2020-04-03
Mq6

sveinbjornt:
* Various additions/improvements/fixes in arithmetic, currency, date, petrol, distance, special, wiki, unit qmodules
* Some fixes/improvements in geo data and geo.py module
* Query stats chart is now a bar chart (makes more sense)
* Lots of additional tests, incl. test w. dummy test data for clear query history api call (we burned ourselves on that one a while back!)
* New scraper for Vísindavefurinn (incomplete and disabled)
* Lots of additions to Vocab.conf


queries/currency.py
```diff
@@ -205,10 +226,17 @@ def help_text(lemma):
...
+    | "kanadadollar" # Common mistake
     | currency_cad/fall
 
+QCurCAD_nf →
+    "kanadadalur" | "kanadadalurinn"
+    | "kanadadollari" | "kanadadollarinn"
+
 QCurCAD_þgf →
-    "kanadadal" | "kanadadalnum"
+    "kanadadal" | "kanadadalnum" | "kanadadollar"
```
vthorsteinsson, line 239: No need to repeat `"kanadadollar"` since it's already been defined in `QCurCAD/fall`

queries/currency.py
```diff
@@ -205,10 +226,17 @@ def help_text(lemma):
 
 QCurCAD/fall →
     | 'kanadískur:lo'_kk/fall 'dollari:kk'/fall
+    | 'kanadadalur:no'_kk/fall
```
vthorsteinsson, line 229: These forms are not in BÍN (but should be!)  Also, it should be `'kanadadalur:kk'/fall`, not `:no`

queries/currency.py
```diff
@@ -285,6 +318,8 @@ def help_text(lemma):
     | QCurConvertAmount QCurMuchIn
     # Hvað eru 10 dollarar mikið [í evrum]?
     | QCurConvertAmount QCurMuch
+    # Hvað fæ ég margar krónur fyrir 10 dollara?
+    # | "hvað" "fæ" "ég" QCurMany "krónur" fyrir
```
vthorsteinsson, line 322: `"fyrir"` should be in double quotes

queries/petrol.py
```diff
@@ -174,6 +174,9 @@ def help_text(lemma):
 QPetrolHere →
     "hér" | "hérna"
 
+QPetrolNow →
+    "í" "dag" | "eins" "og" "stendur" | "í" "augnablikinu" | "þessa" "dagana"
+
```
vthorsteinsson, line 179: Add "núna"?

queries/wiki.py
```diff
@@ -124,17 +125,39 @@ def help_text(lemma):
     | "gætirðu" "flett" "upp" QWikiSubjectÞgf "í" QWikipedia
 
 QWikiSubjectNf →
+    QWikiPrevSubjectNf | QWikiSubjectNlNom
```
vthorsteinsson, line 128: Better to not mix up Icelandic Nf/Þf/Þgf/Ef with English Nom/Acc/Dat/Gen?

queries/currency.py
sveinbjornt: Haha, I was planning to make a few commits before pressing  "Request review" but you've jumped the gun! :P

queries/currency.py
sveinbjornt: True, but commented out :)

queries/wiki.py
sveinbjornt: So true. I'm actually a bit confused about this and I think we should establish strict conventions. So we stick with the Icelandic names of cases? So it's nf, þf, þgf and ef, as opposed to nom, acc, dat, gen?

sveinbjornt: When you think about it, this probably makes the codebase more transparent for Icelandic speakers at the cost of making it more opaque for others. Decisions, decisions!

PR #131, opened 2020-07-02
Tests adapted to abbreviation changes in Tokenizer

Holado:


Holado:
The checks should pass when all the packages have been updated @vthorsteinsson

PR #133, opened 2020-08-26
Homecontrol frá Hinriki

sveinbjornt:


config/Vocab.conf
```diff
@@ -833,6 +833,13 @@ neyðar-sím-kerfi hk
 Sunnefju-foss kk
 þrjú-leyti hk
 fjögur-leyti hk
+axlar-síður lo
```
sveinbjornt, line 836: These additions to Vocab.conf slipped in by accident, cleaner if you could remove them, and the dbclean.py script. They'll be erged into the codebase via another PR.

utils/dbclean.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
```
sveinbjornt, line 1: Remove dbclean.py from this PR.

queries/js/connectHub.js
```diff
@@ -0,0 +1,90 @@
+//const XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;
```
sveinbjornt, line 1: I suggest that you add "use strict" to the top of all the JS documents and run a linter on them (e.g. jshint) to make sure everything is top notch.

queries/home.py
```diff
@@ -0,0 +1,394 @@
...
+    selected_light = device_data['smartlights']['selected_light']
+    hue_credentials = device_data['smartlights'].get('philips_hue')
+
+    # connect smartdevice action
+    if "qtype" in result and result.qtype == "ConnectSmartDevice":
+
+        answer = "Skal gert"
+        host = flask.request.host
+
+        js = read_jsfile("connectHub.js")
```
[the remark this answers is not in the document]
sveinbjornt: Just wrap read_jsfile, e.g.

from functools import lru_cache

@lru_cache(maxsize=32)
def read_jsfile(fn):

db/models.py
```diff
@@ -684,3 +684,26 @@ def __repr__(self):
...
+    # Key to distinguish between different types of json data that can be stored
+    key = Column(String, nullable=False, primary_key=True)
+
+    # created
+    created = Column(DateTime, nullable=False)
+
+    # Last modified
+    last_modified = Column(DateTime, nullable=False)
+
+    # data
```
sveinbjornt, line 708: Capitalize comments

queries/__init__.py
```diff
@@ -623,3 +623,13 @@ def timezone4loc(loc, fallback=None):
     if fallback and fallback in country_timezones:
         return country_timezones[fallback][0]
     return None
+
+
+@lru_cache(maxsize=32)
+def read_jsfile(filename):
+    from rjsmin import jsmin
+    basepath, _ = os.path.split(os.path.realpath(__file__))
+    _JSDIR = basepath + os.sep + "js"
```
vthorsteinsson, line 632: Use `os.path.join()` instead of os.sep

queries/__init__.py
```diff
@@ -623,3 +623,13 @@ def timezone4loc(loc, fallback=None):
...
+    fpath = _JSDIR + os.sep + filename
```
vthorsteinsson, line 633: Ditto

queries/js/light.js
```diff
@@ -0,0 +1,48 @@
...
+
+    let { ipAddress, username } = serviceStorage;
+
+    if (!ipAddress || !username) {
+        return 'Snjalltæki er ekki tengt';
+    }
+    else {
+        try {
+            changeLight(ipAddress, username, on, dimmer);
+            return 'skal gert';
```
vthorsteinsson, line 43: *Skal gert

queries/js/lightInfo.js
```diff
@@ -0,0 +1,82 @@
...
+
+    var request = new XMLHttpRequest();
+    request.open('GET', `http://${ipAddress}/api/${username}/lights`, false);  // `false` makes the request synchronous
+    request.send(null);
+
+    if (request.status === 200) {
+        return JSON.parse(request.responseText);
+    }
+    else {
+        throw new Error('Error while fetchin light info');
```
vthorsteinsson, line 15: *fetching

queries/js/selectLight.js
```diff
@@ -0,0 +1,34 @@
...
+
+    let { ipAddress, username } = serviceStorage;
+
+    if ( !ipAddress || !username) {
+        return 'Snjalltæki er ekki tengt';
+    }
+    else {
+        try {
+            changeLight(ipAddress, username, lightNo);
+            return 'skal gert';
```
vthorsteinsson, line 29: *Skal gert

queries/js/test.js
```diff
@@ -0,0 +1,104 @@
...
+    serviceStorage = {};
+    let deviceInfo = getSmartDeviceAddress();
+
+    try {
+        let username = createNewDeveloper(deviceInfo.internalipaddress);
+        
+        if (username.success){
+            serviceStorage.username = username.success.username;
+        } else {
+            return 'Ýttu á \'Philips\' takkan á tengiboxinu og reyndu aftur';
```
vthorsteinsson, line 91: *takkann

routes/api.py
```diff
@@ -583,3 +583,73 @@ def exit_api():
         raise RuntimeError("Not running with the Werkzeug Server")
     shutdown_func()
     return "The server has shut down"
+
+
+@routes.route("/register_query_data.api", methods=["POST"])
+def register_query_data_api():
```
vthorsteinsson, line 589: Looks like a security hole - anybody could call this with any client id?

db/models.py
```diff
@@ -685,3 +685,29 @@ def __repr__(self):
         return "Feedback(name='{0}', email='{1}', topic='{2}', comment='{3}')".format(
             self.name, self.email, self.topic, self.comment
         )
+
+
+class QueryData(Base):
+
+    __tablename__ = "query_data"
```
vthorsteinsson, line 692: I suggest renaming this to `querydata` (no underscore)

db/models.py
```diff
@@ -685,3 +685,29 @@ def __repr__(self):
...
+
+    __table_args__ = (PrimaryKeyConstraint("client_id", "key", name="querydata_pkey"),)
+
+    client_id = Column(String, nullable=False, primary_key=True)
+
+    # Key to distinguish between different types of json data that can be stored
+    key = Column(String, nullable=False, primary_key=True)
+
```
vthorsteinsson, line 700: Add a module column before the key column

`module = Column(String, nullable=False, primary_key=True)`

vthorsteinsson: ...and there should be a max column size for these string columns, probably 64 characters

queries/js/lightInfo.js
sveinbjornt: Yeah, this branch needs a lot of work to meet our standards :). Will get on it tomorrow.

db/models.py
sveinbjornt: I disagree. Unnecessary paranoid namespace separation. There's no reason to associate any particular key with any particular query module, a key value pair such as username might be of value to multiple modules. Why do we need a qmodule name to disambiguate?

vthorsteinsson: Yes, valid point. Although there might then be a convention for a global namespace, such as module="global". But there could alternatively be a convention for modules to prefix their keys with their name, such as "homecontrol:device" or something like that. - The concern arises because in the future, we may have independent third parties developing modules, and it is not reasonable to expect them to know about each others' key namespaces.

vthorsteinsson:
Looks good to me. OK to merge.

routes/api.py
sveinbjornt: Yes, I'll extend the key functionality implemented for the speech API to this endpoint as well.

PR #135, opened 2020-09-14
Mq10

sveinbjornt:
Lots of minor stuff, but it would be nice to get some of these into master ASAP:

* Removed support for Python 3.5
* Updated package versions in requirements.txt
* Support for "án vasks" queries in arithmetic qmodule
* Manifold improvements in geo, userloc, distance and various other qmodules
* Additions to Vocab.conf


queries/geography.py
```diff
@@ -173,6 +173,7 @@ def QGeoLocationDescQuery(node, params, result):
     (r"norður kóre", "Norður-Kóre"),
     (r"norður kaledón", "Norður-Kaledón"),
     (r"^seychelles.+$", "Seychelles"),
+    (r"^taiwans$", "Taiwan"),
```
vthorsteinsson, line 176: What about 'taíwan' or 'taívan' - does the speech recognizer not return those forms?

queries/special.py
```diff
@@ -71,12 +71,14 @@
     "Þú getur til dæmis beðið mig um upplýsingar úr Wikipedíu.",
     "Þú getur til dæmis beðið mig um að leysa einföld reikningsdæmi.",
     "Þú getur til dæmis spurt mig um mælieiningar.",
-    "Þú getur til dæmis spurt mig hvað er í sjónvarpinu.",
+    "Þú getur til dæmis spurt mig um hvað er í sjónvarpinu.",
```
vthorsteinsson, line 74: "...spurt mig hvað er í sjónvarpinu."

queries/userloc.py
```diff
@@ -56,6 +56,7 @@
 
 QUserLocationCurrent →
     "hvar" "er" "ég" QULocEiginlega? QULocLocated? QULocInTheWorld? QULocNow?
+    | "hvar" QULocInTheWorld? "er" "ég" QULocEiginlega QULocLocated? QULocNow?
```
vthorsteinsson, line 59: The rules in lines 58 and 59 are a bit suspicious-looking - they might be ambiguous, i.e. both would match the same string. That is perhaps not a big problem but we generally try to avoid explicit ambiguity.

config/Vocab.conf
```diff
@@ -874,6 +874,25 @@ eftirlauna-sparnaður kk
 # so: "skala"?
 # sirka?
 
+# Bætum við nýyrðinu 'váhrif'
```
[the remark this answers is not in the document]
vthorsteinsson: ...in the plural, I mean.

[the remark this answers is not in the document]
sveinbjornt: Yeah, that's what I thought too, but BÍN has "áhrif" in the singular! https://bin.arnastofnun.is/beyging/36704

queries/geography.py
sveinbjornt: "taívan" and "tævan" already handled (both in BÍN) but the speech recognizer sometimes returns "taiwan".

PR #136, opened 2020-09-16
Location lookup for US states + more

sveinbjornt:
* Location analysis can now look up info on US states (which in turn improves geography qmodule etc.)
* Refactoring wrt geo resource files
* Various improvements in other qmodules
* Additional tests + refactoring of tests
* Fixes in Fréttablaðið scraper, which now uses their RSS feed and handles "special" article pages (already cherry-picked into master and not viewable in the diff)


sveinbjornt:
A look at this would be appreciated, @vthorsteinsson.

geo.py
```diff
@@ -352,6 +373,52 @@ def lookup_city_info(name):
     return city_lookup(cn)
 
 
+US_STATE_NAMES = None  # type: Optional[Dict[str, str]]
```
vthorsteinsson, line 376: OK to change to `US_STATE_NAMES: Optional[Dict[str, str]] = None`

geo.py
```diff
@@ -352,6 +373,52 @@ def lookup_city_info(name):
...
+
+
+def code_for_us_state(name):
+    """ Given a US state name string, canonical or Icelandicized,
+        return the state's 2-char code. """
+    names = _load_us_state_names()  # Lazy-load
+    return names.get(name.strip())
+
+
+US_STATE_COORDS = None  # type: Optional[Dict[str, str]]
```
vthorsteinsson, line 399: Move type annotation to variable declaration, instead of comment

tests/test_queries.py
```diff
@@ -618,6 +622,10 @@ def test_query_utility_functions():
     assert not is_plural(22.1)
     assert not is_plural(22.41)
 
+    assert sing_or_plur(21, "maður", "menn") == "21 maður"
```
vthorsteinsson, line 625: Add test for 11 and 2.11 and 2.011

PR #137, opened 2020-09-21
Name change f. ReynirPackage

sveinbjornt:
Renaming:

ReynirPackage-->GreynirPackage
Reynir-->Greynir

Changed in almost all files. @vthorsteinsson You might want to take a look at some of the old source files in vectors/* and utils/*, which I mostly left alone.


scripts/deploy.sh
```diff
@@ -43,8 +43,8 @@ cd $DEST
 
 
 echo "Removing binary grammar files"
-rm p369/site-packages/reynir/Reynir.grammar.bin
-rm p369/site-packages/reynir/Reynir.grammar.query.bin
+rm p369/site-packages/reynir/Greynir.grammar.bin
```
vthorsteinsson, line 46: Not ideal to have the virtualenv folder hardcoded like this...

PR #138, opened 2020-09-26
Userinfo

sveinbjornt:
* Query mechanism now remembers user's name if told (stored in QueryData table), and uses it in selected replies
* Type annotation in several root modules and all the query modules


doc.py
```diff
@@ -53,11 +53,11 @@ def __init__(self, path_or_bytes):
             self.data = path_or_bytes
 
     @abc.abstractmethod
-    def extract_text(self):
+    def extract_text(self) -> str:
         """ All subclasses must implement this method """
         pass
 
-    def write_to_file(self, path):
+    def write_to_file(self, path: str):
```
vthorsteinsson, line 60: `def write_to_file(self, path: str) -> None:`

geo.py
```diff
@@ -350,13 +352,13 @@ def location_info(name, kind, placename_hints=None):
...
 
-ICE_CITY_NAMES = None  # type: Optional[Dict[str, str]]
+ICE_CITY_NAMES: Optional[Dict[str, str]] = None
 ICE_CITIES_JSONPATH = os.path.join(
     os.path.dirname(__file__), "resources", "geo", "cities_is.json"
 )
 
 
-def _load_city_names():
+def _load_city_names() -> Dict:
```
vthorsteinsson, line 361: Should probably be `-> Dict[str, str]:`

geo.py
```diff
@@ -414,15 +416,15 @@ def _load_us_state_coords():
     return US_STATE_COORDS
 
 
-def coords_for_us_state_code(code):
+def coords_for_us_state_code(code: str) -> Optional[Tuple]:
```
vthorsteinsson, line 419: Probably should be `-> Optional[Tuple[float, float]]:`

vthorsteinsson: ...or, even better, define `LatLonTuple = Tuple[float, float]` somewhere centrally and use it consistently for (lat, lon) tuples

geo.py
```diff
@@ -740,13 +752,13 @@ def iceprep_for_placename(pn):
...
 
-def iceprep_for_cc(cc):
+def iceprep_for_cc(cc: str) -> str:
     """ Return the right Icelandic preposition ("í" or "á") for
         a country, given its ISO country code, e.g. "IS". """
     return "á" if cc.upper() in CC_ICEPREP_A else "í"
 
 
-def iceprep_for_country(cn):
+def iceprep_for_country(cn: str) -> Optional[str]:
```
vthorsteinsson, line 761: Or just `str`? Does it return an empty string or `None` for unknown countries?

images.py
```diff
@@ -35,8 +37,8 @@
 QUERY_TIMEOUT = 4.0
 
 
-def _server_query(url, q):
-    """ Query a server via HTTP GET with a URL-encoded query string obtained from the dict q """
+def _server_query(url: str, q: Dict) -> Optional[bytes]:
```
vthorsteinsson, line 40: Probably q must be `Dict[str, str]`?

images.py
```diff
@@ -80,7 +82,12 @@ def _server_query(url, q):
 
 
 def get_image_url(
-    name, hints=[], size="large", thumb=False, enclosing_session=None, cache_only=False
+    name: str,
+    hints: List = [],
```
vthorsteinsson, line 86: Might want a `*` in the parameter list before `hints`, demanding that all parameters from here on must be keyword parameters

images.py
```diff
@@ -80,7 +82,12 @@ def _server_query(url, q):
...
+    size: str = "large",
+    thumb: bool = False,
+    enclosing_session=None,
```
vthorsteinsson, line 89: Missing type for `enclosing_session`

queries/bus.py
```diff
@@ -619,7 +620,7 @@ def voice_distance(d):
     return "{0}0 metrar".format(m)
 
 
-def hms_fmt(hms):
+def hms_fmt(hms: Tuple) -> str:
```
vthorsteinsson, line 623: `Tuple[int, int, int]`

queries/counting.py
```diff
@@ -28,15 +28,15 @@
...
+from query import Query
 
 _COUNTING_QTYPE = "Counting"
 
 
 TOPIC_LEMMAS = ["telja"]
 
 
-def help_text(lemma):
+def help_text(lemma: str):
```
vthorsteinsson, line 39: `-> str`

PR #147, opened 2021-05-23
Flight module

sultur:
Answers basic questions about flights arriving to/departing from Icelandic airports.


queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+    This module handles queries relating to air travel.
+
+"""
+
+import re
+import random
+import logging
+import cachetools
+from html import escape
+from typing import List, Dict, Any, Optional
```
vthorsteinsson, line 31: By convention, `from typing import...` should be the first import in the import list.

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+    | QFlightsWhen "mætir"
+    | "hver" "er" "lendingartími"
+    | "hver" "er" "lendingartíminn" "fyrir"
+
+QFlightsDepartureQuery →
+    QFlightsWhenDepNextPlane QFlightsPathDesc
+
+QFlightsWhenDepNextPlane →
+    QFlightsWhen "leggur" QFlightsNextPlane "af" "stað"
+    > QFlightsWhenDep QFlightsNextPlane "af" "stað"
```
vthorsteinsson, line 102: Better: 

```
QFlightsAfStað -> "af" "stað"

...
| QFlightsWhenDep QFlightsNextPlane QFlightsAfStað?
```

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+
+# This module wants to handle parse trees for queries
+HANDLE_TREE = True
+
+# The grammar nonterminals this module wants to handle
+QUERY_NONTERMINALS = {"QFlights"}
+
+# The context-free grammar for the queries recognized by this plug-in module
+GRAMMAR = """
+
```
vthorsteinsson, line 77: Would be good to have comments here with examples of the supported query forms.

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+QFlightsWhen →
+    "hvenær" | "klukkan" "hvað"
+
+$score(-35) QFlightsWhen
+
+QFlightsPlane →
+    "flug" | "flugvél" | "flugvélar" | "vél" | "vélar" | "flugs"
+
+$score(+100) QFlightsPlane
```
vthorsteinsson, line 125: Not sure that these `$score()` pragmas are needed - is there a reason for them?

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+
+QFlightsPathDesc →
+    # Specify at least one endpoint
+    QFlightsPrepLoc QFlightsPrepLoc > QFlightsPrepLoc
```
vthorsteinsson, line 129: Insert line break before `>`, per convention

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+    | "á" QFlightsArrLoc_þgf
+    | "á" QFlightsArrLoc_þf
+
+QFlightsArrLoc/fall →
+    Nl/fall
+
+QFlightsDepLoc/fall →
+    Nl/fall
+
+$tag(keep) QFlightsArrLoc/fall
```
vthorsteinsson, line 144: Well done ;-)

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+    "ifj": "ísafjörður",
+    "kef": "keflavík",
+    "rkv": "reykjavík",
+    "sak": "sauðárkrókur",
+    "tho": "þórshöfn",
+    "vey": "vestmannaeyjar",
+    "vpn": "vopnafjörður",
+}
+
+_AIRPORT_TO_IATA_MAP = {
```
vthorsteinsson, line 173: Better: `{ val: key for key, val in _IATA_TO_AIRPORT_MAP.items() }`

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+    "keflavík": "kef",
+    "reykjavík": "rkv",
+    "sauðárkrókur": "sak",
+    "þórshöfn": "tho",
+    "vestmannaeyjar": "vey",
+    "vopnafjörður": "vpn",
+}
+
+# Day indices in accusative case
+_DAY_INDEX_ACC = {
```
vthorsteinsson, line 191: Is this nowhere else to be found in the code? Could extract to `__init__.py` or a common module for use in other queries.

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+
+_FLIGHTS_CACHE_TTL = 600  # seconds, ttl = 10 mins
+
+# Cache for flights either departing or arriving
+_FLIGHT_CACHE = cachetools.TTLCache(maxsize=2, ttl=_FLIGHTS_CACHE_TTL)
+
+
+def _fetch_flight_data(
+    from_date: datetime, to_date: datetime, iata_code: str, departing: bool
+) -> List[Dict[str, Any]]:
```
vthorsteinsson, line 260: Consider defining a type early in the file: `FlightList = List[Dict[str, Any]]`

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+    flight_info: List[Dict[str, Any]] = []
+    for flight in flights:
+        if n <= 0:
+            break
+
+        if (
+            airport == "*"
+            or airport in flight["DisplayName"].lower()
+            or (
+                flight["AltDisplayName"] and airport in flight["AltDisplayName"].lower()
```
vthorsteinsson, line 309: `flight.get("AltDisplayName")` is perhaps safer here, if the key is missing from the dict?

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+    now: datetime = datetime.now(timezone.utc)  # Timezone aware datetime
+
+    flight_info: List[Dict[str, Any]] = []
+    for flight in flights:
+        if n <= 0:
+            break
+
+        if (
+            airport == "*"
+            or airport in flight["DisplayName"].lower()
```
vthorsteinsson, line 307: Is a string "in" operation definitely OK here, as opposed to for instance `.startswith()` or a straight comparison?

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+            or (
+                flight["AltDisplayName"] and airport in flight["AltDisplayName"].lower()
+            )
+        ):
+            # Use estimated time instead of scheduled if available
+            if flight["Estimated"]:
```
vthorsteinsson, line 313: `flight.get("Estimated")` safer?

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+                flight_time = datetime.fromisoformat(flight["Estimated"])
+            else:
+                flight_time = datetime.fromisoformat(flight["Scheduled"])
+
+            # Make sure flight isn't in the past
+            if flight_time and flight_time >= now:
+                flight["flight_time"] = flight_time
```
vthorsteinsson, line 320: This is a bit gross - you're modifying elements of the original flight list that was passed into the function, in-place. It's probably better to copy the dict items you need to a new dict, which is added to the result list.

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+                    f"Flugi {flight['No']} frá {airport} til {api_airport} er aflýst."
+                )
+            else:
+                line = (
+                    f"Flug {flight['No']} frá {airport} "
+                    f"lendir {prep} {api_airport} {flight_date_str} "
+                    f"klukkan {flight_time_str} að staðartíma."
+                )
+
+        voice_line = re.sub(r" \d+\. ", " " + _DAY_INDEX_ACC[flight_dt.day] + " ", line)
```
vthorsteinsson, line 391: This normalization is great - and would be good to have in a common library function, possibly in `__init__.py` in the queries directory.

queries/flights.py
```diff
@@ -0,0 +1,502 @@
...
+def sentence(state: QueryStateDict, result: Result) -> None:
+    """Called when sentence processing is complete."""
+    q: Query = state["query"]
+    if (
+        "qtype" in result
+        and result["qtype"] == _FLIGHTS_QTYPE
+        and "departure" in result
+    ):
+        try:
+            with changedlocale(category="LC_TIME"):
```
vthorsteinsson, line 491: Is this required? Why?

vthorsteinsson (commented):
Very well done!

tests/test_queries.py
```diff
@@ -367,7 +367,94 @@ def test_query_api(client):
...
+        {"q": "hvenær fer næsta flug til blabla frá ekkitil", "voice": True},
+        "Flights",
+    )
+    assert re.search(no_matching_flight_pattern, json["answer"])
+    json = qmcall(
+        c,
+        {"q": "hvenær fer næsta flug frá ekkitil til blablab", "voice": True},
+        "Flights",
+    )
+    assert re.search(no_matching_flight_pattern, json["answer"])
```
sveinbjornt, line 457: I hesitate to say it, but this is a lot of tests vis-a-vis the tests for the other query modules in test_queries. Our continuous integration testing pipeline is taking long enough as it is, and all these HTTP queries are quite expensive. Are they all necessary? @vthorsteinsson What's our policy on this? Do we want full HTTP comm testing on all the query modules or should we move over to testing the queries directly via code rather than through the API? Seems like the way to go to me.

vthorsteinsson: I see your point, but on the other hand there can almost never be too much testing! I hazard a guess that the HTTP overhead is pretty negligible compared to the actual query processing, so I say we let this stay as-is (and not be too reluctant to add more tests of this nature to other queries).

sveinbjornt: OK, I'll extend the tests, have been reluctant to do so hitherto due to basic performance issues.

queries/flights.py
sultur: Fixed!

queries/flights.py
sultur: Fixed!

queries/flights.py
sultur: Good point! I added a few examples to the grammar, do you think more would be necessary/helpful?

queries/flights.py
sultur: I removed the score pragmas and it had no effect. I added them to try and fix a bug where queries of the form "Hver er brottfarartími ..." were interpreted as queries asking about a person called "brottfarartími ...". I fixed that elsewhere in the grammar and forgot to remove these :).

queries/flights.py
sultur: Fixed!

queries/flights.py
sultur: Fixed!

queries/flights.py
sultur: Fixed!

queries/flights.py
sultur: Very good point, fixed!

queries/flights.py
```diff
@@ -112,21 +122,18 @@ def help_text(lemma: str) -> str:
...
-$score(+100) QFlightsPlane
-
 QFlightsPathDesc →
-    # Specify at least one endpoint
-    QFlightsPrepLoc QFlightsPrepLoc > QFlightsPrepLoc
+    # At least one endpoint of the flight (in any order), e.g.
+    # frá Keflavík til Reykjavíkur
+    # til Akureyrar
+    QFlightsPrepLoc QFlightsPrepLoc
+    > QFlightsPrepLoc
```
vthorsteinsson, line 136: Not totally sure that this is wise; more canonical is simply `QFlightsPrepLoc QFlightsPrepLoc?`, omitting the second production.

queries/flights.py
```diff
@@ -112,21 +122,18 @@ def help_text(lemma: str) -> str:
...
 
-$score(+35) QFlightsWhenDep
-
 QFlightsWhen →
     "hvenær" | "klukkan" "hvað"
 
-$score(-35) QFlightsWhen
-
 QFlightsPlane →
     "flug" | "flugvél" | "flugvélar" | "vél" | "vélar" | "flugs"
```
vthorsteinsson, line 129: If we wanted to be very precise, we would have two functions here: QFlightsPlane_nf and QFlightsPlane_ef. Even more fancy would be:

`QFlightsPlane/fall -> 'flug:hk'_et/fall | 'flugvél:kvk'_et/fall | 'vél:kvk'_et/fall`

This would catch the different grammatical forms automatically, including indefinite or definite ('vél' / 'vélin' etc.)

queries/flights.py
```diff
@@ -258,12 +243,18 @@ def QFlightsDepLoc(node, params, result):
 _FLIGHTS_CACHE_TTL = 600  # seconds, ttl = 10 mins
 
 # Cache for flights either departing or arriving
-_FLIGHT_CACHE = cachetools.TTLCache(maxsize=2, ttl=_FLIGHTS_CACHE_TTL)
+_FLIGHT_CACHE: cachetools.TTLCache = cachetools.TTLCache(
+    maxsize=2, ttl=_FLIGHTS_CACHE_TTL
+)
+
+# For type checking
+FlightType = Dict[str, Any]
```
vthorsteinsson, line 251: FYI, for maximum fanciness and robustness, you could put in a `TypedDict` here. A TypedDict specifies the types of values for each string key in the dict. In this case, there is a limited number of known string keys so a TypedDict would work well.

queries/flights.py
```diff
@@ -323,20 +314,26 @@ def _filter_flight_data(
...
-                flight["api_airport"] = api_airport
 
-                flight_info.append(flight)
+                # Create copy of dictionary and
+                # add flight_time and api_airport attributes
+                flight_copy: FlightType = {
+                    **flight,
+                    "flight_time": flight_time,
+                    "api_airport": api_airport,
+                }
```
vthorsteinsson, line 324: Nicely done!

queries/flights.py
sultur: I originally did that but kept running into the problem of queries such as "Hvenær fer næsta flug frá New York til Keflavík?" only using one QFlightsPrepLoc and interpreting 'New York til Keflavík' as a single location.
Do you have an idea how that could be fixed/circumvented?

vthorsteinsson: OK, I see your point. Strictly speaking we shouldn't be using the `Nl/fall` nonterminal in the main grammar; it should be one of the more restrictive sub-nonterminals underl `Nl` that doesn't allow a preposition after the noun phrase. But your original hack is an OK solution considering the complexity of the alternatives.

queries/flights.py
sultur: That is a lot cleaner. Fixed it, thank you!

queries/flights.py
sultur: It would be very handy to have a function that turns integers into written Icelandic ordinals (for voice lines and such), is there any such functionality somewhere? (Otherwise I could attempt to create one in another pull request)

queries/flights.py
sultur: I added it for good measure.

queries/flights.py
sultur: Added it for good measure.

queries/flights.py
sveinbjornt: No, I've been meaning to implement something along those lines for a while now, but feel free to take a stab at it!

queries/flights.py
sultur: I created a TypedDict but had to ignore a few warnings to get a clean run, as the static type checker sometimes raised false positives (e.g. thinking that a variable could still be None inside an if-statement that explicitly checked for it not being None).

vthorsteinsson: In my experience the type checkers are still primitive in this regard; I usually resort to going back from `x["y"]` to `x.get("y", "")`, where the field `y` is an `Optional[str]`; this silences the warnings.

queries/flights.py
```diff
@@ -385,8 +385,10 @@ def _format_flight_answer(flights: FlightList) -> Dict[str, str]:
 
     for flight in flights:
         assert isinstance(flight.get("No"), str)
+        assert isinstance(flight.get("DisplayName"), str)
```
vthorsteinsson, line 388: These should not be needed, with judicious use of .get() instead of []

sultur: okidoki, fixed!

queries/flights.py
```diff
@@ -0,0 +1,543 @@
...
+    q: Query = state["query"]
+    if (
+        "qtype" in result
+        and result["qtype"] == _FLIGHTS_QTYPE
+        and isinstance(result.get("departure"), bool)
+    ):
+        try:
+            answ: Dict[str, str] = _process_result(result)
+            q.set_qtype(_FLIGHTS_QTYPE)
+            q.set_answer(answ, answ["answer"], answ["voice"])
```
sveinbjornt, line 535: You might want to set the source for flight information, (e.g. q.set_source("Isavia"))

sultur: Added source, thanks!

queries/flights.py
```diff
@@ -0,0 +1,543 @@
...
+    _FLIGHT_CACHE[departing] = res["Items"]
+    return res["Items"]
+
+
+def _attribute_airport_match(flight: FlightType, attribute: str, airport: str) -> bool:
+    """
+    Safely checks whether the string flight[attribute] in lowercase
+    either starts or ends with airport string.
+    """
+    return isinstance(flight.get(attribute), str) and (
```
vthorsteinsson, line 306: Better:
```
   a = flight.get(attribute)
   if not isinstance(a, str):
      return False
   a = a.lower()
   return a.startswith(airport) or a.endswith(airport)

```

sultur: fixed!

PR #150, opened 2021-05-28
Minor format() fixes in tel and ja query modules

sultur:
Call to format() in tel.py always inserted "None". Added missing format() in ja.py so output isn't "{0} er ekki gilt símanúmer".


sveinbjornt:
Good catch!

PR #152, opened 2021-06-01
Radio schedules

sultur:
Added Rás 1 and Rás 2 schedule querying to queries/schedules.py.
Minor changes to TV schedule functionality to prepare better for multiple channels.
(Was going to add querying for other TV channels than RÚV but the endpoints don't seem to work on apis.is)

### DIFF
--- a/queries/schedules.py
+++ b/queries/schedules.py
@@ -118,6 +118,7 @@ QScheduleRadio →
 
 QSchRadioStationNowQuery →
     QSchWhatIsNom QSchEiginlega? QSchGoingOn? QSchOnRadioStation QSchNow?
+    | QSchWhatIsNom QSchEiginlega? QSchOnSchedule? QSchOnRadioStation QSchNow?
     | QSchWhatIsDative QSchEiginlega? "verið" "að" "spila" QSchOnRadioStation QSchNow?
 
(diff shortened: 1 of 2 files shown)


queries/schedules.py
```diff
@@ -25,6 +25,7 @@
 """
 
 # TODO: Support TV schedule queries for other stations than RÚV
+#       (the endpoints currently seem to be broken on apis.is, as of June 1st 2021)
```
sveinbjornt, line 28: Might want to take a look at how the apis.is guys were fetching this info? Maybe it makes more sense to talk directly to the source instead of proxying through apis.is?

queries/schedules.py
```diff
@@ -49,11 +50,14 @@
...
     "sjónvarp",
     "sjónvarpsdagskrá",
+    "útvarpsdagskrá",
     "dagskrá",
     "rúv",
     "ríkissjónvarp",
     "útvarp",
     "ríkisútvarp",
+    "rás",
+    "þáttur",
```
sveinbjornt, line 60: "útvarpsrás", "útvarpsþáttur"

sultur: done!

sultur:
Working on better pull request on main repo.

PR #154, opened 2021-06-25
Test and qmodule improvements

sveinbjornt:
* Massive refactoring of query module tests. Now each module has its own test function (cleaner).
* Tweaks in ja, tel qmodules
* Better handling of errors in currency qmodule + switch over to using arion source for exchange rates (lb broken)
* Various minor tweaks/improvements

PR #157, opened 2021-07-02
Currency endpoint fix

sultur:
Small fix which changes currency endpoint from lb to m5 (Landsbankinn API endpoint returns error code 500), along with Black formatting fixes in tests.


sultur:
Pull request #154 addresses the same issue.

PR #158, opened 2021-07-02
Flight module fixes

sultur:
Removed datetime.fromisoformat call from flight module, which might have been cause of errors in tests for Python version 3.6 as it is Python 3.7+ specific.


vthorsteinsson (commented):
LGTM

PR #162, opened 2021-08-11
Grammar for parsing numbers in text form

sultur:
Grammar idea for parsing spoken numbers.
Should probably be incorporated in Greynir.grammar under a specific flag (hopefully simplified if possible).

### DIFF
--- a/queries/text_to_num.py
+++ b/queries/text_to_num.py
@@ -294,11 +294,7 @@ QNum0 →
     'núll'
 
 QNum1Til9 →
-    # 'einn'/fall/tala/kyn  # FIXME: Catch different cases of 1,2,3 and 4
-    # | 'tveir'/fall/tala/kyn
-    # | 'þrír'/fall/tala/kyn
-    # | 'fjórir'/fall/tala/kyn
-    to/tala/fall/kyn
+    to/tala/fall/kyn  # Numbers 1-4
     | "fimm"
     | "sex"
     | "sjö"
@@ -371,12 +367,11 @@ QNumOktilljón →
 
 
 """
-# TODO: CATCH CASES AND GENDER OF NUMBERS !!! Bæta inn í Vocab.conf töluorðum fyrir tölur ekki í BÍN
+# TODO: CATCH CASES AND GENDER OF NUMBERS !!! Add to Vocab.conf numbers not in BÍN
 
-# TODO: "oktilljón og milljón septilljónir" og "septilljón og milljón sextilljónir",
-#       Spyrja hvort eigi að leyfa eða ekki t.d. milljarður septilljóna == eh margar oktilljónir í staðinn
-
-# TODO: "tólf hundruð þúsund"?
+# TODO: "oktilljón og milljón septilljónir" & "septilljón og milljón sextilljónir",
+#       Allow "milljarður septilljóna" == "... oktilljónir"?
+#       Allow "tólf hundruð þúsund"?
 # TODO: Ordinals
 # TODO: "á annan tug manna"? "fjórir tugir"
 
@@ -421,7 +416,7 @@ _NUMBERS = {
     "trilljón": 10 ** 18,
     "trilljarður": 10 ** 21,
     "kvaðrilljón": 10 ** 24,
-    # TODO: Add to BÍN
+    # Following numbers not in BÍN
     "kvaðrilljarður": 10 ** 27,
     "kvintilljón": 10 ** 30,
     "sextilljón": 10 ** 36,
@@ -695,6 +690,8 @@ def sentence(state: QueryStateDict, result: Result) -> None:
     q.set_answer(*gen_answer(answer))
     q.set_qtype(_NUM_QTYPE)
 
+
+# TODO: BÍN additions in format for Vocab.conf
 
 # kvaðrilljarður    kvaðrilljarður        kk alm NFET
 # kvaðrilljarður    kvaðrilljarð          kk alm ÞFET
--- a/queries/text_to_num.py
+++ b/queries/text_to_num.py
@@ -1,0 +1,801 @@
+"""
+
+    Greynir: Natural language processing for Icelandic
+
+    Opinion query response module
+
+    Copyright (C) 2021 Miðeind ehf.
+
+       This program is free software: you can redistribute it and/or modify
+       it under the terms of the GNU General Public License as published by
+       the Free Software Foundation, either version 3 of the License, or
+       (at your option) any later version.
+       This program is distributed in the hope that it will be useful,
+       but WITHOUT ANY WARRANTY; without even the implied warranty of
+       MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+       GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see http://www.gnu.org/licenses/.
+
+
+    Matches a single number in written form and returns its value.
+
+"""
+
+from typing import Sequence
+from query import Query, QueryStateDict
+from tree import Result, Node
+
+from queries import gen_answer
+
+
+_NUM_QTYPE = "Number"
+
+
+# Indicate that this module wants to handle parse trees for queries,
+# as opposed to simple literal text strings
+HANDLE_TREE = True
+
+# The grammar nonterminals this module wants to handle
+QUERY_NONTERMINALS = {"QNumQuery"}
+
+# The context-free grammar for the queries recognized by this plug-in module
+GRAMMAR = """
+
+Query →
+    QNumQuery
+
+QNumQuery →
+    QNumber '?'?
+
+QNumber →
+    QNumUndirOktilljón | QNumOktilljónir QNumUndirOktilljón? | QNumOktilljónir QNumOgUndirOktilljón
+
+#################
+# NON-TERMINALS #
+#################
+
+#### OKTILLJÓNIR ####
+
+QNumOktilljónir →
+    QNumUndirOktilljón? QNumOktilljón
+
+#### SEPTILLJÓNIR ####
+
+QNumUndirOktilljón →
+    QNumSeptilljónir QNumUndirSeptilljón?
+    | QNumSeptilljónir QNumOgUndirSeptilljón
+    | QNumUndirSeptilljón
+
+QNumOgUndirOktilljón →
+    "og" QNumSeptilljónir
+    | QNumOgUndirSeptilljón
+
+QNumSeptilljónir →
+    QNumUndirSeptilljón? QNumSeptilljón
+
+#### SEXTILLJÓNIR ####
+
+QNumUndirSeptilljón →
+    QNumSextilljónir QNumUndirSextilljón?
+    | QNumSextilljónir QNumOgUndirSextilljón
+    | QNumUndirSextilljón
+
+QNumOgUndirSeptilljón →
+    "og" QNumSextilljónir
+    | QNumOgUndirSextilljón
+
+QNumSextilljónir →
+    QNumUndirSextilljón? QNumSextilljón
+
+#### KVINTILLJÓNIR ####
+
+QNumUndirSextilljón →
+    QNumKvintilljónir QNumUndirKvintilljón?
+    | QNumKvintilljónir QNumOgUndirKvintilljón
+    | QNumUndirKvintilljón
+
+QNumOgUndirSextilljón →
+    "og" QNumKvintilljónir
+    | QNumOgUndirKvintilljón
+
+QNumKvintilljónir →
+    QNumUndirKvintilljón? QNumKvintilljón
+
+#### KVAÐRILLJARÐAR ####
+
+QNumUndirKvintilljón →
+    QNumKvaðrilljarðar QNumUndirKvaðrilljarði?
+    | QNumKvaðrilljarðar QNumOgUndirKvaðrilljarði
+    | QNumUndirKvaðrilljarði
+
+QNumOgUndirKvintilljón →
+    "og" QNumKvaðrilljarðar
+    | QNumOgUndirKvaðrilljarði
+
+QNumKvaðrilljarðar →
+    QNumUndirKvaðrilljarði? QNumKvaðrilljarður
+
+#### KVAÐRILLJÓNIR ####
+
+QNumUndirKvaðrilljarði →
+    QNumKvaðrilljónir QNumUndirKvaðrilljón?
+    | QNumKvaðrilljónir QNumOgUndirKvaðrilljón
+    | QNumUndirKvaðrilljón
+
+QNumOgUndirKvaðrilljarði →
+    "og" QNumKvaðrilljónir
+    | QNumOgUndirKvaðrilljón
+
+QNumKvaðrilljónir →
+    QNumUndirKvaðrilljón? QNumKvaðrilljón
+
+#### TRILLJARÐAR ####
+
+QNumUndirKvaðrilljón →
+    QNumTrilljarðar QNumUndirTrilljarði?
+    | QNumTrilljarðar QNumOgUndirTrilljarði
+    | QNumUndirTrilljarði
+
+QNumOgUndirKvaðrilljón →
+    "og" QNumTrilljarðar
+    | QNumOgUndirTrilljarði
+
+QNumTrilljarðar →
+    QNumUndirTrilljarði? QNumTrilljarður
+
+#### TRILLJÓNIR ####
+
+QNumUndirTrilljarði →
+    QNumTrilljónir QNumUndirTrilljón?
+    | QNumTrilljónir QNumOgUndirTrilljón
+    | QNumUndirTrilljón
+
+QNumOgUndirTrilljarði →
+    "og" QNumTrilljónir
+    | QNumOgUndirTrilljón
+
+QNumTrilljónir →
+    QNumUndirTrilljón? QNumTrilljón
+    | QNumUndirMilljarði? QNumMilljarður QNumMilljarður
+
+#### BILLJARÐAR ####
+
+QNumUndirTrilljón →
+    QNumBilljarðar QNumUndirBilljarði?
+    | QNumBilljarðar QNumOgUndirBilljarði
+    | QNumUndirBilljarði
+
+QNumOgUndirTrilljón →
+    "og" QNumBilljarðar
+    | QNumOgUndirBilljarði
+
+QNumBilljarðar →
+    QNumUndirBilljarði? QNumBilljarður
+    | QNumUndirMilljón? QNumMilljón QNumMilljarður
+
+#### BILLJÓNIR ####
+
+QNumUndirBilljarði →
+    QNumBilljónir QNumUndirBilljón?
+    | QNumBilljónir QNumOgUndirBilljón
+    | QNumUndirBilljón
+
+QNumOgUndirBilljarði →
+    "og" QNumBilljónir
+    | QNumOgUndirBilljón
+
+QNumBilljónir →
+    QNumUndirBilljón? QNumBilljón
+    | QNumUndirMilljón? QNumMilljón QNumMilljón
+
+#### MILLJARÐAR ####
+
+QNumUndirBilljón →
+    QNumMilljarðar QNumUndirMilljarði?
+    | QNumMilljarðar QNumOgUndirMilljarði
+    | QNumUndirMilljarði
+
+QNumOgUndirBilljón →
+    "og" QNumMilljarðar
+    | QNumOgUndirMilljarði
+
+QNumMilljarðar →
+    QNumUndirMilljarði? QNumMilljarður
+
+#### MILLJÓNIR ####
+
+QNumUndirMilljarði →
+    QNumMilljónir QNumUndirMilljón?
+    | QNumMilljónir QNumOgUndirMilljón
+    | QNumUndirMilljón
+
+QNumOgUndirMilljarði →
+    "og" QNumMilljónir
+    | QNumOgUndirMilljón
+
+QNumMilljónir →
+    QNumUndirMilljón? QNumMilljón
+
+#### ÞÚSUND ####
+
+QNumUndirMilljón →
+    QNumÞúsundir QNumUndirÞúsund?
+    | QNumÞúsundir QNumOgUndirÞúsund
+    | QNum10Til19Hundruð QNumUndirHundrað?
+    | QNum10Til19Hundruð QNumOgUndirHundrað?
+    | QNumUndirÞúsund
+
+QNumOgUndirMilljón →
+    "og" QNumÞúsundir
+    | QNumOgUndirÞúsund
+
+QNumÞúsundir →
+    QNumUndirÞúsund? QNumÞúsund
+
+QNum10Til19Hundruð →
+    QNum10Til19 QNumHundrað
+
+#### HUNDRUÐ ####
+
+QNumUndirÞúsund →
+    QNumHundruð QNumTugurOgEining?
+    | QNumHundruð QNumOgUndirHundrað?
+    | QNumUndirHundrað
+
+QNumOgUndirÞúsund →
+    "og" QNumHundruð
+    | QNumOgUndirHundrað
+
+QNumHundruð →
+    QNum1Til9? QNumHundrað
+
+#### UNDIR HUNDRAÐ ####
+
+QNumUndirHundrað →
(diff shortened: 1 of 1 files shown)


sultur:
Note: This only works if GreynirPackage doesn't join together numbers (such as "fimm hundruð þúsund") into a single token.

query.py
```diff
@@ -437,9 +439,23 @@ def init_class(cls) -> None:
         cls._tree_processors = [t[1] for t in sorted(tree_procs, key=lambda x: -x[0])]
         cls._text_processors = [t[1] for t in sorted(text_procs, key=lambda x: -x[0])]
 
-        # Obtain query grammar fragments from the tree processors
+        # Query grammar plug-ins:
+        # Additional grammar fragments,
+        # which can be used in query module grammars
+        cls._query_grammar_additions = []
+        plugnames = modules_in_dir("queries/plugins")
```
sultur, line 446: (We may want to define the path "queries/plugins" as a constant somewhere, I just hardcoded it for the time being as this is only an idea.)

queries/currency.py
```diff
@@ -35,6 +35,7 @@
 
 from query import Query, QueryStateDict
 from queries import query_json_api, iceformat_float, sing_or_plur, gen_answer
+from queries.plugins.text_to_num import *
```
vthorsteinsson, line 38: We tend to avoid import *, according to PEP guidelines. The code is more readable if all imports are explicit.

queries/plugins/text_to_num.py
```diff
@@ -0,0 +1,573 @@
...
+
+
+    Nonterminal for a single number written in natural language.
+    Returns its value in the list result["numbers"].
+
+"""
+
+# TODO: Add to Vocab.conf numbers not in BÍN (see bottom of file)
+# TODO: Deal with cases and genders of numbers (/fall/kyn/tala)
+# TODO: Allow "tólf hundruð þúsund"? Allow "milljón milljónir"?
```
vthorsteinsson, line 29: Yes, both "tólf hundruð þúsund" and "milljón milljónir" should preferably be allowed.

queries/plugins/text_to_num.py
```diff
@@ -0,0 +1,573 @@
...
+# TODO: Ordinals (TöluðRaðtala)
+# TODO: "fjórir tugir", "tylft" & "á annan tug manna"?
```
vthorsteinsson, line 31: These should however not be included.

queries/plugins/text_to_num.py
```diff
@@ -0,0 +1,573 @@
...
+
+from typing import Sequence
+from query import QueryStateDict
+from tree import Result, Node
+
+# The context-free grammar for the queries recognized by this plug-in module
```
vthorsteinsson, line 37: Update comment - these are not "queries". Probably enough to say "The context-free grammar for number utterances recognized by this utility module."

queries/plugins/text_to_num.py
```diff
@@ -0,0 +1,573 @@
...
+TöluðTalaMilljónir →
+    TöluðTalaUndirMilljón? TöluðTalaMilljón
+
+# ÞÚSUND
+
+TöluðTalaUndirMilljón →
+    TöluðTalaÞúsundir TöluðTalaUndirÞúsund?
+    | TöluðTalaÞúsundir TöluðTalaOgUndirÞúsund
+    | TöluðTala10Til19Hundruð TöluðTalaUndirHundrað?
+    | TöluðTala10Til19Hundruð TöluðTalaOgUndirHundrað?
```
vthorsteinsson, line 215: These two productions are mutually ambiguous. Refine the grammar so that only one production can match a token string that doesn't end with "TöluðTalaUndirHundrað" or "TöluðTalaOgUndirHundrað".

queries/plugins/text_to_num.py
```diff
@@ -0,0 +1,573 @@
...
+    TöluðTalaUndirÞúsund? TöluðTalaÞúsund
+
+TöluðTala10Til19Hundruð →
+    TöluðTala10Til19 TöluðTalaHundrað
+
+# HUNDRUÐ
+
+TöluðTalaUndirÞúsund →
+    TöluðTalaHundruð TöluðTalaTugurOgEining?
+    | TöluðTalaHundruð TöluðTalaOgUndirHundrað?
```
vthorsteinsson, line 232: Same here, these two are mutually ambiguous.

queries/plugins/text_to_num.py
```diff
@@ -0,0 +1,573 @@
...
+    "tuttugu"
+    | "þrjátíu"
+    | "fjörutíu"
+    | "fimmtíu"
+    | "sextíu"
+    | "sjötíu"
+    | "áttatíu"
+    | "níutíu"
+
+TöluðTalaHundrað → 'hundrað'
```
vthorsteinsson, line 295: Missing :cat (word category specifier)? Also be careful that in plural, there are both 'hundruð' and 'hundruðir' (and 'þúsund' and 'þúsundir').

queries/plugins/text_to_num.py
```diff
@@ -0,0 +1,573 @@
...
+
+TöluðTalaÞúsund → 'þúsund:töl'
+
+TöluðTalaMilljón → 'milljón'
```
vthorsteinsson, line 299: :cat ?

queries/plugins/text_to_num.py
```diff
@@ -0,0 +1,573 @@
...
+    "þúsund": 1000,
+    "milljón": 10 ** 6,
+    "miljarður": 10 ** 9,
+    "milljarður": 10 ** 9,
+    "billjón": 10 ** 12,
+    "billjarður": 10 ** 15,
+    "trilljón": 10 ** 18,
+    "trilljarður": 10 ** 21,
+    "kvaðrilljón": 10 ** 24,
+    # Following numbers not in BÍN
```
vthorsteinsson, line 368: Maybe understandably ;-)

queries/plugins/text_to_num.py
```diff
@@ -0,0 +1,573 @@
...
+    "sextilljón": 10 ** 36,
+    "septilljón": 10 ** 42,
+    "oktilljón": 10 ** 48,
+}
+
+
+# (math.prod is only available for python 3.8+)
+def _prod(li: Sequence[int]) -> int:
+    """Take product of non-empty list."""
+    p = li[0]
```
vthorsteinsson, line 380: Cool Python one-liner: `return functools.reduce(operator.__mul__, li, 1)`  (requires import functools, import operator)

queries/plugins/text_to_num.py
```diff
@@ -0,0 +1,573 @@
...
+    TöluðTalaBilljarðar,
+    TöluðTalaTrilljónir,
+    TöluðTalaTrilljarðar,
+    TöluðTalaKvaðrilljónir,
+    TöluðTalaKvaðrilljarðar,
+    TöluðTalaKvintilljónir,
+    TöluðTalaSextilljónir,
+    TöluðTalaSeptilljónir,
+    TöluðTalaOktilljónir,
+) = [_multiply_children] * 15
```
vthorsteinsson, line 410: Cool!

query.py
vthorsteinsson: Yes. a constant at the beginning of the file would be better. Also, double-check that the forward slash is OK here, as this may potentially run on Windows which uses backslashes. A safe way is to use `os.path.join("queries", "plugins")`.

query.py
```diff
@@ -437,9 +439,23 @@ def init_class(cls) -> None:
...
+                cls._query_grammar_additions.append(m)
+            except ImportError as e:
+                logging.error(
+                    "Error importing query plugin {0}: {1}".format(plug, e)
+                )
+
+        # Obtain query grammar fragments from the tree processors and query plugins
         grammar_fragments: List[str] = []
-        for processor in cls._tree_processors:
+        for processor in cls._tree_processors + cls._query_grammar_additions:
```
vthorsteinsson, line 458: Might be better to have the query grammar additions in front of the tree processors. I think the grammar parser handles forward references OK, but nevertheless it is logical to include them in that order.

queries/plugins/text_to_num.py
vthorsteinsson: Actually plain `operator.mul` works, the dunder underscores are not needed.

queries/currency.py
sultur: I import the functions like this so they get called in `NonterminalNode.process` in `tree.py` (so they aren't prefixed with `text_to_num.`). Is there some cleaner way to make sure the `TöluðTala...` functions get called for each appropriate nonterminal? (Having an explicit import in this case means importing about 48 functions, which is a bit much).

I could e.g. include the `text_to_num` module in the TreeStateDict `state`, and search for the relevant function there, if it isn't found in `processor`. Or is there some way to do this with fancy imports instead?

vthorsteinsson: OK, I understand. Then the best solution is just to put a comment before the import explaining what is happening, i.e. that these are nonterminal handler functions that are deliberately being imported into the parent namespace.

queries/plugins/text_to_num.py
sultur: Added category specifiers! We may need to add hundruðir as an entry to ord.add.csv - as it is relatively common (but wrong, according to BÍN). Added þúsundir:kvk for plural.

sultur:
Large improvements to this pull request :), but it is not quite ready yet though. I would like to move repeated code from the query modules into the `queries/utility` folder, but I don't have time to do that yet as I'm prioritizing the IoT PR.

tests/test_queries.py
```diff
@@ -310,9 +310,21 @@
     json = qmcall(client, {"q": "hvað eru tíu þúsund krónur margir dalir"}, "Currency")
     assert re.search(r"^\d+(,\d+)?$", json["answer"]) is not None
 
+    json = qmcall(client, {"q": "hvað eru þrettán hundruð fjörutíu og sex krónur margir dalir"}, "Currency")
+    assert re.search(r"^\d+(,\d+)?$", json["answer"]) is not None
+
+    json = qmcall(client, {"q": "hvað eru átján milljónir króna margir dalir"}, "Currency")
+    assert re.search(r"^(\d?\d?\d\.?)+(,\d+)?$", json["answer"]) is not None
```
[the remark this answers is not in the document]
sultur: :expressionless:

sultur:
Should we merge this pull request, **after a review**?

## Pros
It contains very helpful functionality for other pull requests (which could make them cleaner and more manageable) and currently doesn't break any functionality in the query modules or elsewhere that I know of.

The STT normalizes numbers for us at the moment, so the new tokenization and number parsing isn't needed for the query modules usual behaviour.

Also it doesn't require any radical changes to query module creation or Greynir's usage, despite the wall of text which I'm currently writing.

## Cons?
It still has room for improvements, mainly specifying better the inflections of numbers in the grammar and extracting more functionality from other query modules than `currency.py`, but that could be done in other pull requests so this one doesn't become extremely large.

# Main changes summarized
## `queries/utility/`
Created the `queries/utility` folder which should house shared functionality between the query modules. This required changes to `query.py`, `tree.py` and minor changes to `processor.py`.

### `query.py`
Functions imported from tree query modules are now kept in a `ChainMap` instead of a module object (the chain map contains nonterminal functions from the utility modules, along with functions from the imported module). So `Query._tree_processors` is now a list of `ChainMap` objects.

The chain maps prefer the imported processor's functions over the shared utility functions, so authors of query modules can overwrite utility module functionality on a module-by-module basis if needed.

### `tree.py`
I modified some functions to accomodate the the chain maps instead of module objects, which mostly just involved changing `getattr()` calls to `.get()` calls.

I created a type called `ProcEnv` (processing environment), which is just a type describing a mapping of keywords/nonterminal names to functions (a `ChainMap` is a mapping, so it's also a `ProcEnv`). Functions which used to accept module objects now accept `ProcEnv` objects.

### `processor.py`
The aforementioned changes however also meant I had to pass in `vars(p)` in some places in `processor.py` instead of just `p` (in order to change the module objects into mappings of functions to match the `ProcEnv` type). The functionality stays the same there though. Also added type hints and calls to `vars()` in `test_processors.py` where necessary.

## Query tokenization
Incoming queries **_(note: only used in query processing)_** are now tokenized using `no_multiply_numbers=True`. This means written numbers, e.g. _"tvö hundruð"_ are no longer a single token, but rather two word tokens which should be parsed using the the `UHeilTala` nonterminal (fractions can be parsed using `UBrotaTala`, see `queries/currency.py` and `tests/test_queries.py::test_currency` for examples of these).

## `pathlib.Path` instead of `os.path`
`util.py` now has `GREYNIR_ROOT_PATH: pathlib.Path` available as a variable (found using `LICENSE.txt` in the root of the repo). This makes it easy (and cleaner in my opinion) to reference any file path in the Greynir repo (platform agnostic too). Refactored a bunch of usages of `os.path.join` to utilize this instead, while keeping the functionality the same. This also allowed removing of a bunch of ~~hacky~~ functions that manipulated the path in order to find certain files.

queries/__init__.py
```diff
@@ -52,17 +52,17 @@
 from pytz import country_timezones
 
 from geo import country_name_for_isocode, iceprep_for_cc, LatLonTuple
-from queries.num import number_to_text, float_to_text
+from queries.utility.number_utils import number_to_text, float_to_text
```
sveinbjornt, line 55: Suggestion: queries/utility -> queries/util and queries.utility.number_utils -> queries.util.num or queries.util.numbers
Good to keep naming nice and short and non-redundant.

sultur: Good point, I'll shorten the names :)

sultur: This causes confusion with `util.py` in the root. Is it ok if I rename that file to `utility.py`?

sveinbjornt: Does it, though? The root level util is global, whereas queries.util is utilities for queries. Seems pretty clear to me. But do what you feel is best. The main point is to keep nested hierarchies, files and folder names short to avoid horrible Java-style deep nested hierarchies with excessively verbose names.

main.py
```diff
@@ -31,12 +31,14 @@
...
-from typing import Callable, Dict, Pattern, Optional, Union
+from typing import Dict, List, Pattern, Optional, Union
 
 import sys
-import os
 import re
 import logging
+from os import environ as ENV
+from platform import system as os_name
+from pathlib import Path
```
sveinbjornt, line 41: I approve! Pathlib was introduced in Python 3.4 and is a much cleaner way of working with paths. Agree that the os.path stuff is ugly and hacky.

queries/utility/numbers_grammar.py
```diff
@@ -0,0 +1,184 @@
...
+    Exposes nonterminal "UHeilTala" for parsing
+    numbers either written in natural language or in digits,
+    along with "UBrotaTala" for parsing floats ("(number) 'komma' (numbers)")
+    written in natural language.
+    Returns the number values in the list result["numbers"].
+
+"""
+
+# TODO: Deal better with cases and genders of numbers
+# TODO: Allow "tólf hundruð þúsund" & "hundruðir" (need to add hundruðir to ord.add/auka.csv)
```
sveinbjornt, line 32: "hundruðir" is technically wrong (see http://malfar.arnastofnun.is/grein/65992) but I guess we should allow it? It's a very common error.

sultur: Yeah, I was thinking that too.

sveinbjornt: Good thing you got me thinking about this. Turns out Yfirlestur doesn't catch "hundruðir"!

Screen Shot 2022-10-27 at 14 13 39

queries/__init__.py
sultur: It at least confuses the linter when I do `from util import GREYNIR_ROOT_PATH, read_api_key` in `queries/__init__.py` because we always import as if were in the root, but the util folder is also in the queries folder. I also tested importing from `..util` but then it complains about not having a parent module :/.

queries/currency.py
```diff
@@ -21,14 +21,14 @@
 
 """
 
+# TODO: Switch from using apis.is
```
sveinbjornt, line 24: I agree. That project is dead, hasn't seen any love in over 4 years, and apparently is not being maintained based on the accumulation of unresolved issues and ignored pull requests.

We should be able to talk directly Arion, Landsbankinn or Seðlabankinn (probably best).

The apis.is code is here, should contain the API endpoint URLs we need:
https://github.com/apis-is/apis/tree/master/endpoints/currency

queries/__init__.py
sveinbjornt: OK, name it as you like, we wouldn't want to confuse the linter :).

tree.py
```diff
@@ -44,14 +45,14 @@
...
+from typing_extensions import Required, TypedDict
 
-from types import ModuleType
 
 import json
 import re
 
-from collections import OrderedDict
 import abc
+from collections import OrderedDict
```
sveinbjornt, line 55: Actually, we should get rid of all OrderedDicts in the codebase and replace them with regular dicts. As of Python 3.7, all Python dicts are ordered, and we now only support 3.7+.

util.py
```diff
@@ -20,23 +20,53 @@
...
-import os
 from functools import lru_cache
+from pathlib import Path
+
+
+def _find_greynir_root_folder() -> Path:
+    """Small helper function to find the root folder of Greynir."""
+    p = Path(__file__).resolve()
+    # Search for LICENSE.txt file (should be in root folder)
+    while not (p / "LICENSE.txt").is_file() and p.parent != p:
```
sveinbjornt, line 33: I dunno, looking for LICENSE.txt feels a bit hacky, but I can't think of a good alternative method off the top of my head.

tree.py
sultur: Ah interesting, I didn't add that import, I think I just moved it to be above the imports from our packages to make the imports look nicer. We can address the OrderedDict changes in a separate PR.

util.py
```diff
@@ -20,23 +20,53 @@
...
     try:
-        with open(path) as f:
-            return f.read().strip()
+        return p.read_text().strip()
     except FileNotFoundError:
         pass
     return ""
 
 
+def modules_in_dir(*args: str) -> List[str]:
```
sveinbjornt, line 55: Good, this belongs in util! Might want to take a look at speech/__init__.py, which contains an unfortunate duplicate of this function, if I remember correctly.

util.py
sultur: Hmm, yeah, I couldn't think of another file that was guaranteed to be in our root and not change names.

util.py
sultur: Nice, I'll remove the `speech/__init__.py` implementation of `modules_in_dir` :+1:

sveinbjornt:
Looks good to me! But before merging, it'd be great if @vthorsteinsson took a look as well.

queries/__init__.py
sultur: Shortened some of the file/folder names :+1:.

util.py
sultur: I guess we could change to searching for the parent folder being named "Greynir"? (I would need to rename the folder for my fork from "GreynirFork" though)

sveinbjornt: No, just leave it as it is. The name of the repo is more likely to change in the future than LICENSE.txt.

tree.py
vthorsteinsson: Yes, be a little bit careful about this one - I vaguely remember having had a reason to use OrderedDict.

util.py
sultur: The most important idea for this function was that running it from anywhere within the repo should always return the same path (although I think it only gets called from the root directory anyway when the server is run normally).

processor.py
```diff
@@ -265,7 +251,7 @@ def go_single(self, url: str) -> None:
                             ptype: str = getattr(p, "PROCESSOR_TYPE")
                             assert ptype in _PROCESSOR_TYPES, "Unknown processor type"
                             if ptype == _PROCESSOR_TYPE_TREE:
-                                tree.process(session, p)
+                                tree.process(session, vars(p))
```
vthorsteinsson, line 254: Is it still logical to have the `token_container.process()` call below using the old scheme (p instead of vars(p))?

util.py
sultur: hm ok I can try using `__file__` instead :)

queries/util/number_grammar.py
```diff
@@ -0,0 +1,184 @@
...
+    UTöluðTalaUndirBilljarði,
+    UTöluðTalaUndirTrilljón,
+    UTöluðTalaUndirTrilljarði,
+    UTöluðTalaUndirKvaðrilljón,
+    UTöluðTalaUndirKvaðrilljarði,
+    UTöluðTalaUndirKvintilljón,
+    UTöluðTalaUndirSextilljón,
+    UTöluðTalaUndirSeptilljón,
+    UTöluðTalaUndirOktilljón,
+) = [_sum_children] * 15
```
vthorsteinsson, line 154: Clever!

query.py
```diff
@@ -626,13 +678,17 @@ def execute_from_tree(self) -> bool:
...
+                    self._session,
+                    processor,
+                ):
                     # This processor found an answer, which is already stored
                     # in the Query object: return True
                     return True
             except Exception as e:
                 logging.error(
-                    f"Exception in execute_from_tree('{processor.__name__}') "
+                    f"Exception in execute_from_tree('{processor['__name__']}') "
```
vthorsteinsson, line 691: Is `__name__` definitely in the dictionary? (Despite starting with an underscore.)

tests/test_queries.py
```diff
@@ -310,9 +310,46 @@ def test_currency(client: FlaskClient) -> None:
     json = qmcall(client, {"q": "hvað eru tíu þúsund krónur margir dalir"}, "Currency")
     assert re.search(r"^\d+(,\d+)?$", json["answer"]) is not None
 
+    json = qmcall(
+        client,
+        {"q": "hvað eru þrettán hundruð fjörutíu og sex krónur margir dalir"},
+        "Currency",
+    )
+    assert re.search(r"^\d+(,\d+)?$", json["answer"]) is not None
```
vthorsteinsson, line 318: Consider adding a check that the number is not only present, but also correct?

tree.py
```diff
@@ -62,12 +63,16 @@
...
+# Processing environment
+# A mapping of keywords or nonterminal names to functions
+# for processing sentence trees or, more specifically, query trees
+ProcEnv = MutableMapping[str, Any]
 
 
 class TreeStateDict(TypedDict, total=False):
-    session: Session
-    processor: ModuleType
+    session: Required[Session]
```
vthorsteinsson, line 74: Does this definitely work from Python 3.7 onwards? You don't need to import from typing_extensions?

utility.py
```diff
@@ -20,23 +20,53 @@
...
     try:
-        with open(path) as f:
-            return f.read().strip()
+        return p.read_text().strip()
     except FileNotFoundError:
         pass
     return ""
 
 
+def modules_in_dir(*args: str) -> List[str]:
```
sveinbjornt, line 55: Maybe it would be a good idea to write some minor tests for this function? Could be added to tests/test_util.py

tree.py
sultur: I import `Required` and `TypedDict` from `typing_extensions` :) I work on PyPy 7.3.9 which corresponds to Python 3.7.13, so this should be all good.

processor.py
sultur: Not really, I think it would make most sense to have the module types all be mappings or `ProcEnv`s. I'll add some fixes.

query.py
sultur: Yes, `__name__` should always be imported from the tree processors (the underscore attributes are only ignored in the `util` modules. I'm gonna change this to a `.get` call though, just to be 100% sure we don't get an exception while logging another exception.

tests/test_queries.py
sultur: currency.py currently doesn't return the amount asked for in its answer, so I added tests that compared the answer to the same question where the number wasn't written in natural language :+1:.

utility.py
sultur: Added some tests :+1:.

util.py
sultur: `__file__` works and is cleaner (I misunderstood how the variable worked), made this much cleaner :+1:.

sultur:
I think I've addressed all your comments, thanks for the review! :smile_cat: Ok to merge?

PR #165, opened 2021-09-01
Schedules additions

sultur:
Schedules module now answers queries for TV (RÚV and Stöð 2 stations) and radio channels (Rás 1/2).


queries/schedules.py
```diff
@@ -90,271 +95,831 @@ def help_text(lemma: str) -> str:
...
+    result["channel_type"] = "tv"
+    result["station"] = "RÚV"
 
-_RUV_RADIO_SCHEDULE_API_ENDPOINT = "https://muninn.ruv.is/files/xml/{0}/{1}/"
-_RADIO_API_ERRMSG = "Ekki tókst að sækja útvarpsdagskrá."
-_RADIO_SCHED_CACHE: Dict[str, Any] = {}
-_RADIO_LAST_FETCHED: Dict[str, datetime] = {}
 
+def QSchRUV2(node: Node, params: ParamList, result: Result) -> None:
+    result["api_channel"] = "ruv2"
```
sveinbjornt, line 327: Maybe write separate utility function to add these attributes to the Result object to reduce code?

queries/schedules.py
```diff
@@ -90,271 +95,831 @@ def help_text(lemma: str) -> str:
...
+    if len(sched) > 0:
+        _SCHED_CACHE[(channel, date)] = sched
+
+    return sched
+
+
+def _get_program_start_end(
+    program: dict, station: str
+) -> Tuple[datetime.datetime, datetime.datetime]:
+    """Return the time span of a episode/program."""
```
sveinbjornt, line 513: "an" episode.

queries/schedules.py
```diff
@@ -90,271 +95,831 @@ def help_text(lemma: str) -> str:
...
+        return []
+    else:
+        url: str = _STATION_ENDPOINTS[station].format(channel, date.isoformat())
+    response = cast(Optional[List], query_json_api(url))
+
+    if response is None:
+        return []
+
+    sched: _SchedType
+    if station == "RÚV":
```
sveinbjornt, line 496: Might want to define canonical station names (e.g. "RÚV" string) as constants instead of doing comparisons repeatedly w. bare string.

queries/schedules.py
```diff
@@ -90,271 +95,831 @@ def help_text(lemma: str) -> str:
...
+
+    # Fetch schedule data from API or cache.
+    sched: _SchedType = _query_schedule_api(api_channel, station, qdt.date())
+
+    if len(sched) == 0:
+        date: str = ""
+        # Add date if not asking about today
+        if qdt.date() != now_date:
+            with changedlocale(category="LC_TIME"):
+                date = qdt.strftime(" %-d. %B")
```
sveinbjornt, line 871: Use your new icestrftime function here instead of changelocale?

queries/schedules.py
```diff
@@ -90,271 +95,831 @@ def help_text(lemma: str) -> str:
...
+            station=station,
+            channel=channel,
+            expire_time=qdt,
+        )
+
+    curr_prog: Optional[dict]
+    next_prog: Optional[dict]
+    curr_prog, next_prog = _get_current_and_next_program(sched, station, qdt, get_next)
+
+    with changedlocale(category="LC_TIME"):
```
sveinbjornt, line 888: Ditto.

query.py
```diff
@@ -143,7 +143,7 @@ def __call__(
 )
 _IGNORED_PREFIX_RE = r"^({0})\s*".format("|".join(_IGNORED_QUERY_PREFIXES))
 # Auto-capitalization corrections
-_CAPITALIZATION_REPLACEMENTS = (("í Dag", "í dag"),)
+_CAPITALIZATION_REPLACEMENTS = (("í Dag", "í dag"),("Á dagskrá", "á dagskrá"))
```
sveinbjornt, line 146: What problem exactly is this addressing?

queries/schedules.py
```diff
@@ -90,271 +95,831 @@ def help_text(lemma: str) -> str:
...
+            answer += " " + _clean_desc(next_desc)
+
+        # Answer is valid for station until the next program starts
+        prog_endtime, _ = _get_program_start_end(next_prog, station)
+    else:
+        answer = f"Það er ekkert á dagskrá á {channel} eftir núverandi dagskrárlið."
+
+    voice = answer
+    return {
+        "response": {"answer": answer, "voice": voice},
```
sveinbjornt, line 749: "voice" not needed here.

query.py
sultur: This might be fixed in recent versions of GreynirPackage, I'll test and see, but the issue was simply that "á" was capitalized.

queries/schedules.py
sultur: The icestrftime function I created doesn't work because of some error when calling datetime.strftime with a string containing icelandic characters. It might be a pypy specific thing.

sveinbjornt:
Looks good to me, apart from minor nitpicks. Let me know when this is good to merge.

queries/schedules.py
sultur: Fixed!

queries/schedules.py
sultur: Good point, added a helper function which cleans up the code.

queries/schedules.py
sultur: Oops, fixed

queries/schedules.py
sultur: Added constants for frequently used strings.

query.py
sultur: Removed it, seems to be addressed in new GreynirPackage versions.

PR #171, opened 2021-12-22
Media

sveinbjornt:
* Added support for image_url key in query JSON response. Support for displaying the image is already present in both Embla mobile clients
* New pic qmodule ("Sýndu mér mynd af X?" etc.) using image lookup via images.py (Google Images API)
* New play qmodule that searches YouTube for music and opens video/playlist URL ("Spilaðu lag með Rolling Stones", "Spilaðu góða klassíska tónlist", etc.)
* Wiki module now takes precedence when it comes to "Hvað er X?" queries. Fallback to builtin if no Wikipedia entry for entity in question
* Logging of GPS coordinates for queries has been disabled
* A variety of improvements in other qmodules
* Added /upload_speech_audio.api endpoint (currently does not save data received, effectively disabled)
* Various minor fixes and source formatting


queries/pic.py
```diff
@@ -0,0 +1,177 @@
...
+    q.set_qtype(result.qtype)
+
+
+_WRONG_PIC_ANSWER = "Ég biðst afsökunar á því. Enginn er fullkominn."
+
+
+def _gen_wrong_pic_answer(result: Result, q: Query):
+    """Query states that previous picture was wrong."""
+    q.set_qtype(result.qtype)
+    q.set_answer(*gen_answer(_WRONG_PIC_ANSWER))
```
vthorsteinsson, line 158: We could actually use the query context to tell us which picture is being referred to, and mark it as wrong in the database (or collect this information in some other way for review or use).

queries/play.py
```diff
@@ -0,0 +1,372 @@
+"""
+
+    Greynir: Natural language processing for Icelandic
+
+    Weather query response module
```
vthorsteinsson, line 5: Wrong module name in comment

queries/play.py
```diff
@@ -0,0 +1,372 @@
...
+
+from query import Query
+from queries import gen_answer
+from util import read_api_key
+
+
+_PLAY_QTYPE = "Play"
+
+
+_AFFIRMITIVE = "Skal gert!"
```
vthorsteinsson, line 43: Spelling: _AFFIRMATIVE

queries/play.py
```diff
@@ -0,0 +1,372 @@
...
+
+def _play_electronic(qs: str, q: Query, matches: Optional[Match[str]]) -> None:
+    # Orbital - The Box
+    fb = "https://www.youtube.com/watch?v=qddG0iUSax4"
+    q.set_url(rand_yt_playlist_for_genre("retro electronic music", fallback=fb))
+    q.set_key(matches.group(1) if matches else "")
+
+
+# Play music from a randomly selected genre
+def _play_music(qs: str, q: Query, matches: Match[str]) -> None:
```
vthorsteinsson, line 153: matches parameter is not used, is that intentional? Also, probably should be Optional[Match[str]] if present.

queries/play.py
```diff
@@ -0,0 +1,372 @@
...
+    "spilaðu bíómynd fyrir mig": _play_film,
+    "spilaðu mynd fyrir mig": _play_film,
+    "sýndu kvikmynd": _play_film,
+    "sýndu bíómynd": _play_film,
+    "sýndu mér kvikmynd": _play_film,
+    "sýndu mér bíómynd": _play_film,
+}
+
+_VERB = "|".join(
+    frozenset(
```
vthorsteinsson, line 259: frozenset() is not necessary here, a tuple suffices

queries/play.py
```diff
@@ -0,0 +1,372 @@
...
+
+    # Check if query matches regexes supported by this module
+    matches = None
+    for rx, fn in REGEX_Q2H.items():
+        matches = re.search(rx, ql)
+        if matches:
+            fn(ql, q, matches)
+            break
+
+    if not matches:
```
vthorsteinsson, line 364: This could actually be an else: clause on the for statement. Just sayin'

queries/special.py
```diff
@@ -227,6 +227,16 @@ def _poetry(qs: str, q: Query) -> AnswerType:
     }
 
 
+_STORY = """Einu sinni var karl og kerling í koti.
```
vthorsteinsson, line 230: var -> voru

queries/userinfo.py
```diff
@@ -260,7 +264,9 @@ def _whatsmyaddr_handler(q: Query, ql: str) -> bool:
     r"heimilisfang mitt er (.+)$",
 )
 
-_ADDR_LOOKUP_FAIL = "Ég fann ekki þetta heimilisfang."
+_ADDR_LOOKUP_FAIL = "Ég fann ekki þetta heimilisfang í staðfangaskrá."
+_ADDR_CLIENT_ID_MISSING = "Ég get ekki vistað heimilisfangið því \
```
vthorsteinsson, line 268: Better to have two strings enclosed by parentheses, rathern than a backslash line continuation

routes/api.py
```diff
@@ -659,3 +661,36 @@ def register_query_data_api(version: int = 1) -> Response:
...
+def upload_speech_audio(version: int = 1) -> Response:
+    """Receives uploaded speech audio for a query."""
+
+    # This is disabled for now
+    return better_jsonify(valid=False, errmsg="Not implemented")
+
+    if not (1 <= version <= 1):
+        return better_jsonify(valid=False, errmsg="Unsupported version")
+
+    file = request.files.get("file")
```
vthorsteinsson, line 680: Maybe comment this as debugging/development code?

queries/play.py
sveinbjornt: I don't like else clauses in for statements :)

queries/pic.py
sveinbjornt: Good idea.

PR #175, opened 2022-05-18
Pronunciation of flight numbers, type hints

sultur:
- Voice answer now spells out flight number digits (in kk) and letters,
instead of reading gibberish.
- Added 'Match[str]' type hints in num.py (pylance was complaining).


sultur:
I also made a few modifications to the code but no changes in functionality (there shouldn't be at least).

PR #181, opened 2022-06-09
Split query grammars into separate files (in grammars folder)

sultur:
Idea for (optionally) keeping query module grammars in separate folder.
This means we can write grammar with syntax highlighting (very helpful) and, if needed, split the grammar into multiple files.

I only did two modules to start with as an example,
but I can move all the query modules into separate grammar files if you like the idea :)


queries/yulelads.py
```diff
@@ -139,174 +139,7 @@ def help_text(lemma: str) -> str:
...
-    | "hver" "er" "síðasti" "jólasveinninn"
-    | "hver" "er" "síðastur" "jólasveinanna"
-
-QYuleLad →
-    QYuleLadByDate
-    | QYuleLadFirst QYuleSuffix?
-    | QYuleLadLast QYuleSuffix?
-
-""".format(
+GRAMMAR = read_grammar_file("yulelads").format(
```
sveinbjornt, line 142: I would suggest that the read_grammar_file handle any potential formatting.

queries/date.py
```diff
@@ -136,307 +143,7 @@ def help_text(lemma: str) -> str:
...
-QDateSummerSolstice/fall →
-    'sumarsólstöður:kvk'_ft/fall
-
-QDateWinterSolstice/fall →
-    'vetrarsólstöður:kvk'_ft/fall
-
-$score(+55) QDate
-
-"""
+GRAMMAR = read_grammar_file("date")
```
sveinbjornt, line 146: I don't know if this is a good idea, but maybe we should support automatically loading grammar file with same name as the file where read_grammar_file fn is called? Something like this:

https://stackoverflow.com/questions/59004195/python-how-to-get-the-absolute-path-of-file-where-a-function-was-called#59004672

But I don't know. Explicit is better than implicit acc. Zen of Python.

queries/yulelads.py
sultur: Good point, would something like this work? (fargs here meaning format args)
```
def read_grammar_file(filename: str, *fargs, **fkwargs) -> str:
    ...
    return grammar.format(*fargs, **fkwargs)
```
This would allow us to use either index-based formatting like `"{0}".format("something")` or keyword-based formatting `"{replace_me}".format(replace_me="something")`.

queries/date.py
sultur: We could call the function with something like `__name__` as an argument I guess?

queries/yulelads.py
sultur: See changes in commit [caee77d](https://github.com/mideind/Greynir/pull/181/commits/caee77d598b79a13e15cf4af2e2b7f00fec24c68). I added the possibility of specifying format string substitutions as keyword arguments. (and added arithmetic example, showing a bit more complex formatting than yulelads did)

sveinbjornt: As far as I can tell we have three potential strategies when deciding the directory structure accompanying this change.

The first one is your proposed solution, all grammar files in one folder:

queries/module.py
queries/grammar/module.grammar

The second is keeping module and grammar in the same folder:

queries/module.py
queries/module.grammar

The third is keeping each module in its own directory, which contains files required by the module:

queries/module/module.py
queries/module/module.grammar
queries/module/module.js (or whatever)

The second and third solution have the advantage of containing the grammar file next to the module in hierarchical view, but they're both a bit cluttered. The third solution has its merits, but it makes for a more nested hierarchy which can be a PITA. What do you think, @sultur ?

sveinbjornt:
NB: The read_grammar_file() function will of course need tests before anything goes into master.

queries/yulelads.py
sultur: I think the second solution would be a bit too cluttered. The third has some benefits, maybe when developing more complex query modules (back-and-forth dialogue maybe), but as you mention the nesting can be a bit tiresome. Therefore I think the first solution is good for the time being.
If we want to later change to the third solution we would also have to modify `query.py` to look for folders correctly, but it wouldn't be too much of a hassle.

queries/date.py
sultur: I think it's probably best to be explicit in this case and just provide the filename like so `read_grammar_file("module")`.

PR #195, opened 2022-11-30
Small temporary fix for wrong declension in person titles

sultur:
Small fix to person titles which had words in incorrect cases, such as 'lektor við Háskóli Íslands' (now becomes 'lektor við Háskóla Íslands').
(I only call it temporary because later I would like to refactor some of the the processor code to better utilize handy functions from GreynirPackage.)

### DIFF
--- a/processors/persons.py
+++ b/processors/persons.py
@@ -78,6 +78,8 @@
 import re
 from datetime import datetime
 
+from reynir import NounPhrase
+
 from db.models import Person
 
 from queries import QueryStateDict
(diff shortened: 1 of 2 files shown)


sultur:
@sveinbjornt @vthorsteinsson Do you think this fix might negatively affect other titles?
(I couldn't think of any examples.)
I simply added a call to `NounPhrase(titill).nominative` when a persson is added in `sentence()` in `processors/persons.py`.
If a title isn't parseable, it should just fall back to the raw text, which shouldn't impact weird unparseable titles (but maybe we don't want those anyway?).

sveinbjornt:
LGTM!

vthorsteinsson:
LGTM too :-)

PR #196, opened 2022-11-30
Qr2

sveinbjornt:
* Adds a new, API-key-protected route: /stats/queries containing a dashboard with various charts and stats about query engine usage (no. queries broken down by day, query type distribution, client types, top unanswered and answered queries, etc.)
* Better tests for speech synthesis functionality
* Tweaks and improvements in speak.py
* Various minor fixes and some refactoring


routes/stats.py
```diff
@@ -175,48 +171,196 @@ def stats() -> Union[Response, str]:
...
-            gtotal=gtotal,
-            authresult=authresult,
-            scraped_chart_data=json.dumps(chart_data["scraped"]),
-            parsed_chart_data=json.dumps(chart_data["parsed"]),
-            queries_chart_data=json.dumps(chart_data["queries"]),
-            scraped_avg=int(round(chart_data["scraped"]["avg"])),
-            parsed_avg=round(chart_data["parsed"]["avg"], 1),
-            queries_avg=round(chart_data["queries"]["avg"], 1),
+    try:
+        with SessionContext(read_only=True) as session:
```
vthorsteinsson, line 175: Here we use read_only=True

routes/stats.py
```diff
@@ -175,48 +171,196 @@ def stats() -> Union[Response, str]:
...
+    # Query types
+    res = QueryTypesQuery.period(
+        start=start,
+        end=end,
+        enclosing_session=session,
+    )
+
+    # This function is used to ensure that all the query
+    # types have a fixed, unique color on the pie chart.
+    def _hexcolor4string(s: str) -> str:
```
vthorsteinsson, line 266: Quite a hacky solution. A better one would generate an HSL tuple with constant S (saturation) and L (luminance) but different H (hues), then convert it to RGB.

routes/stats.py
```diff
@@ -175,48 +171,196 @@ def stats() -> Union[Response, str]:
...
+        "datasets": [
+            {
+                "data": [k[0] for k in res],
+                "backgroundColor": [_hexcolor4string(k[1]) for k in res],
+            }
+        ],
+    }
+
+    # Client types
+    _CLIENT_COLORS = {
```
vthorsteinsson, line 288: Again, a good way to allocate these is to do a HSL value with constant S and L but with the Hs equally spaced around a 360 deg circle (and converted to 0..255).

routes/stats.py
```diff
@@ -175,48 +171,196 @@ def stats() -> Union[Response, str]:
...
+        },
+        "query_types": query_types_data,
+        "client_types": client_types_data,
+        "top_unanswered": top_unanswered,
+        "top_answered": top_answered,
+    }
+
+
+@routes.route("/stats/queries", methods=["GET"])
+@cache.cached(timeout=30 * 60, key_prefix="stats", query_string=True)
```
vthorsteinsson, line 335: Do you need both `@cache.cached` and `@max_age`?

routes/words.py
```diff
@@ -128,7 +129,7 @@ def _words2str(words):
 
 def _desc4word(wc):
```
vthorsteinsson, line 130: Missing type annotation

routes/words.py
```diff
@@ -128,7 +129,7 @@ def _words2str(words):
...
     """Create a human-friendly description string for a word/category tuple."""
-    return "{0} ({1})".format(wc[0], CAT_DESC.get(wc[1]))
+    return f"{wc[0]} ({CAT_DESC.get(wc[1])})"
```
vthorsteinsson, line 132: Might be more robust to have a default value for the .get()

routes/stats.py
sveinbjornt: Max age tells the browser how long to cache, whereas the cache decorator enables server-side caching.

db/sql.py
```diff
@@ -165,6 +165,84 @@ def period(
...
+
+
+class TopUnansweredQueriesQuery(_BaseQuery):
+    """Return list of the most frequent *unanswered* queries
+    over a given time period."""
+
+    _Q = """
+        select question, answer, count(question) as qoccurrence from queries
+            where answer is NULL
+            group by question, answer
```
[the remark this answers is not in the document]
sveinbjornt: > ...also, this should probably be the querylog table and not the queries table? queries only has about 12k queries in it, but querylog has ~600k.

The idea is that the dashboard shows the last 30 days of metrics. The QueryLog table contains years' worth of unanswered queries, many of which are now handled.

routes/stats.py
sveinbjornt: Spoken like someone who has addressed a similar problem before! Will do, cap'n! :)

routes/stats.py
sveinbjornt: Nah, these I want hardcoded to reflect the standard color scheme assigned to the different platforms, i.e. light blue for iPhone and the green color for Android.

routes/stats.py
sveinbjornt: Speaking of which, I vaguely remembering solving a similar problem while developing a game for Gogogic in 2010. The point was to create random uniquely coloured shapes that were perceptually different. I used some weird perceptual color space, as I recall. Maybe I still have that code around. That being said, maybe we're overengineering the in-house pie chart.
 ;)

PR #197, opened 2022-12-07
Normalized and improved `bus.py` module

sultur:
Normalization of voice output to `bus.py` (Strætó) module, which also involved:
 - `bus.py` module uses NounPhrase to inflect bus stop names, shortening code. The grammar now uses `UHeilTala`.
 - Fixed `UHeilTala` nonterminal (now also correctly catches digits).
 - Improved engine-specific normalization process (now uses classes).
 - Improved regexes in `numbers_to_text/ordinal`, `floats_to_text`.
 - Implemented/Improved some normalization functions (such as the time handler and email handler).
 - Also added a bunch of tests, along with fixing the needlessly complicated `schedules.py` regex-based tests (they shouldn't fail randomly anymore).

### DIFF
--- a/queries/bus.py
+++ b/queries/bus.py
@@ -36,33 +36,32 @@
 
 # TODO: Hvar er nálægasta strætóstoppistöð?
 # TODO: Hvað er ég lengi í næsta strætóskýli?
+# TODO: Hvar stoppar sjöan næst?
 
-from typing import Iterable, Optional, List, Set, Tuple, Union, cast
+from typing import Dict, Iterable, Optional, List, Set, Tuple, cast
 
+import re
+import random
 from threading import Lock
 from functools import lru_cache
 from datetime import datetime
-import random
 
-from islenska.basics import BinEntry
+from reynir import NounPhrase
 
-import queries
-from queries import AnswerTuple, Query, QueryStateDict, ResponseType, Session
-from tree import Result, Node
+from queries import AnswerTuple, Query, QueryStateDict
+from tree import Result, Node, ParamList
 from queries.util import (
+    is_plural,
     natlang_seq,
     cap_first,
     gen_answer,
     read_grammar_file,
-    sing_or_plur,
 )
-from speech.norm.num import floats_to_text, numbers_to_text
+from speech.norm import gssml, strip_markup
 from settings import Settings
-from reynir import correct_spaces
 from geo import in_iceland
 
-import straeto  # type: ignore  # TODO
-
+import straeto
 
 # Today's bus schedule, cached
 SCHEDULE_TODAY: Optional[straeto.BusSchedule] = None
@@ -148,44 +147,44 @@ GRAMMAR = read_grammar_file("bus")
 # tree processing (depth-first, i.e. bottom-up navigation).
 
 
-def QBusNearestStop(node: Node, params: QueryStateDict, result: Result) -> None:
+def QBusNearestStop(node: Node, params: ParamList, result: Result) -> None:
     """Nearest stop query"""
     result.qtype = "NearestStop"
     # No query key in this case
     result.qkey = ""
 
 
-def QBusStop(node: Node, params: QueryStateDict, result: Result) -> None:
+def QBusStop(node: Node, params: ParamList, result: Result) -> None:
     """Save the word that was used to describe a bus stop"""
     result.stop_word = _WRONG_STOP_WORDS.get(result._nominative, result._nominative)
 
 
-def QBusStopName(node: Node, params: QueryStateDict, result: Result) -> None:
+def QBusStopName(node: Node, params: ParamList, result: Result) -> None:
     """Save the bus stop name"""
     result.stop_name = result._nominative
 
 
-def QBusStopThere(node: Node, params: QueryStateDict, result: Result) -> None:
+def QBusStopThere(node: Node, params: ParamList, result: Result) -> None:
     """A reference to a bus stop mentioned earlier"""
     result.stop_name = "þar"
 
 
-def QBusStopToThere(node: Node, params: QueryStateDict, result: Result) -> None:
+def QBusStopToThere(node: Node, params: ParamList, result: Result) -> None:
     """A reference to a bus stop mentioned earlier"""
     result.stop_name = "þangað"
 
 
-def EfLiður(node: Node, params: QueryStateDict, result: Result) -> None:
+def EfLiður(node: Node, params: ParamList, result: Result) -> None:
     """Don't change the case of possessive clauses"""
     result._nominative = result._text
 
 
-def FsMeðFallstjórn(node: Node, params: QueryStateDict, result: Result) -> None:
+def FsMeðFallstjórn(node: Node, params: ParamList, result: Result) -> None:
     """Don't change the case of prepositional clauses"""
     result._nominative = result._text
 
 
-def QBusNoun(node: Node, params: QueryStateDict, result: Result) -> None:
+def QBusNoun(node: Node, params: ParamList, result: Result) -> None:
     """Save the noun used to refer to a bus"""
     # Use singular, indefinite form
     # Hack: if the QBusNoun is a literal string, the _canonical logic
@@ -197,7 +196,7 @@ def QBusNoun(node: Node, params: QueryStateDict, result: Result) -> None:
     result.bus_noun = result._canonical
 
 
-def QBusArrivalTime(node: Node, params: QueryStateDict, result: Result) -> None:
+def QBusArrivalTime(node: Node, params: ParamList, result: Result) -> None:
     """Bus arrival time query"""
     # Set the query type
     result.qtype = "ArrivalTime"
@@ -208,7 +207,7 @@ def QBusArrivalTime(node: Node, params: QueryStateDict, result: Result) -> None:
         result.qkey = result.bus_number
 
 
-def QBusAnyArrivalTime(node: Node, params: QueryStateDict, result: Result) -> None:
+def QBusAnyArrivalTime(node: Node, params: ParamList, result: Result) -> None:
     """Bus arrival time query"""
     # Set the query type
     result.qtype = "ArrivalTime"
@@ -216,7 +215,7 @@ def QBusAnyArrivalTime(node: Node, params: QueryStateDict, result: Result) -> No
     result.qkey = result.bus_number = "Any"
 
 
-def QBusWhich(node: Node, params: QueryStateDict, result: Result) -> None:
+def QBusWhich(node: Node, params: ParamList, result: Result) -> None:
     """Buses on which routes stop at a given stop"""
     # Set the query type
     result.qtype = "WhichRoute"
@@ -240,53 +239,10 @@ _BUS_WORDS = {
     "tía": 10,
     "ellefa": 11,
     "tólfa": 12,
-    "einn": 1,
-    "eitt": 1,
-    "tveir": 2,
-    "tvö": 2,
-    "þrír": 3,
-    "þrjú": 3,
-    "fjórir": 4,
-    "fjögur": 4,
-    "fimm": 5,
-    "sex": 6,
-    "sjö": 7,
-    "níu": 9,
-    "tíu": 10,
-    "ellefu": 11,
-    "tólf": 12,
-    "þrettán": 13,
-    "fjórtán": 14,
-    "fimmtán": 15,
-    "sextán": 16,
-    "sautján": 17,
-    "átján": 18,
-    "nítján": 19,
-    "tuttugu": 20,
-    "þrjátíu": 30,
-    "fjörutíu": 40,
-    "fimmtíu": 50,
-    "sextíu": 60,
-    "sjötíu": 70,
-    "áttatíu": 80,
-    "níutíu": 90,
-    # Hack to catch common ASR error
-    "hundrað eitt": 101,
-    "hundrað tvö": 102,
-    "hundrað þrjú": 103,
-    "hundrað fjögur": 104,
-    "hundrað fimm": 105,
-    "hundrað sex": 106,
-    "hundrað og eitt": 101,
-    "hundrað og tvö": 102,
-    "hundrað og þrjú": 103,
-    "hundrað og fjögur": 104,
-    "hundrað og fimm": 105,
-    "hundrað og sex": 106,
 }
 
 
-def QBusWord(node: Node, params: QueryStateDict, result: Result) -> None:
+def QBusWord(node: Node, params: ParamList, result: Result) -> None:
     """Handle buses specified in single words,
     such as 'tvisturinn' or 'fimman'"""
     # Retrieve the contained text (in nominative case)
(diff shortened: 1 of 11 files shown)


queries/bus.py
```diff
@@ -301,81 +257,151 @@ def QBusWord(node: Node, params: QueryStateDict, result: Result) -> None:
...
-    np = straeto.BusStop.voice(np)
-    return queries.to_accusative(np, filter_func=_filter_func)
+def accusative_form(np: str, voice: bool = False) -> str:
+    """
+    Return accusative case of the stop name,
+    optionally expanding abbreviations for speech synthesis.
+    """
+    if voice:
+        np = _compass_directions(np)
+        np = _BETTER_NAMES.get(np, np)
```
sveinbjornt, line 315: This logic already exists in voicify_stop_name, abstract it to own fn?

queries/bus.py
```diff
@@ -301,81 +257,151 @@ def QBusWord(node: Node, params: QueryStateDict, result: Result) -> None:
...
-    np = straeto.BusStop.voice(np)
-    return queries.to_dative(np, filter_func=_filter_func)
+def dative_form(np: str, voice: bool = False) -> str:
+    """
+    Return dative case of the stop name,
+    optionally expanding abbreviations for speech synthesis.
+    """
+    if voice:
+        np = _compass_directions(np)
+        np = _BETTER_NAMES.get(np, np)
```
sveinbjornt, line 349: Ditto

queries/bus.py
```diff
@@ -301,81 +257,151 @@ def QBusWord(node: Node, params: QueryStateDict, result: Result) -> None:
...
+
+    split_symb = None
+    if " / " in np:
+        split_symb = " / "
+    elif " - " in np:
+        split_symb = " - "
+
+    if split_symb is None:
```
sveinbjornt, line 357: Isn't this logic also duplicated? Maybe add an abstract fn to handle this split handling logic

queries/bus.py
```diff
@@ -504,35 +518,41 @@ def query_arrival_time(query: Query, session: Session, result: Result):
...
             # "Einarsnesi eða Einarsnesi/Bauganesi"
-            stops_list = " eða ".join([to_dative(s) for s in stops_canonical])
-            answer = (
-                " ".join(["Leiðir", route_seq, "stoppa á", stops_list])
-                + ". Spurðu um eina þeirra."
+            stops_list = natlang_seq([dative_form(s) for s in sorted(stops_canonical)])
+            va_stops_list = " eða ".join(
+                [dative_form(s, voice=True) for s in sorted(stops_canonical)]
             )
+            answer = f"Leiðir {route_seq} stoppa á {stops_list}. Spurðu um eina þeirra."
```
sveinbjornt, line 530: I know you didn't introduce this problem, and it is minor, but maybe worth adding as a TODO at the top of the document: Some bus stops will take the preposition "á" while others use "í". Ideally, this should be handled well ... we have some logic in geo.py if I remember correctly.

speech/__init__.py
```diff
@@ -138,6 +144,7 @@ def normalize(self, voice_string: str) -> str:
         assert (
             len(self._str_stack) == 1
         ), "Error during parsing, are all markup tags correctly closed?"
+        logging.info("Voice string after normalization:", self._str_stack[0])
```
sveinbjornt, line 147: This is quite noisy, isn't it? Don't really think we need to log this.

speech/norm/__init__.py
```diff
@@ -86,7 +85,7 @@ def gssml(data: Any = None, *, type: str, **kwargs: Union[str, int, float]) -> s
     "d": "dé",
     "ð": "eð",
     "e": "e",
-    "é": "je",
+    "é": "é",
```
sveinbjornt, line 88: yeah! ;)

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
+        ac = after_christ is not None and after_christ == "True"
+        return year_to_text(txt, after_christ=ac)
+
+    @classmethod
+    def years(cls, txt: str, *, after_christ: Optional[str] = None):
+        """Voicify text containing multiple years."""
+        ac = after_christ is not None and after_christ == "True"
+        return years_to_text(txt, after_christ=ac)
+
+    # Pronounciation of character names in Icelandic
```
sveinbjornt, line 326: *pronUnciation* :) Common mistake.

tests/test_queries.py
```diff
@@ -263,7 +264,80 @@ def test_bus(client: FlaskClient) -> None:
...
+    )
+    assert json["answer"]
+    assert all(not c.isdecimal() for c in json["voice"])
+
+    json = qmcall(
+        client,
+        {
+            "q": "hvaða leiðir stoppa í Bíó Paradís?",
+            "voice": True,
+            # Note: This adds a row to the QueryLog table
```
sveinbjornt, line 298: Why private=False? Could use test=True param to prevent logging.

sveinbjornt, line 298: :( We don't want the tests to leave anything permanent in the database

sveinbjornt:
Apart from some minor nitpicking, this looks pretty good to me! Wanna take a look, @vthorsteinsson ?

queries/bus.py
sultur: Done! Also fixed a small issue with `NounPhrase` not being able to inflect certain stop names with the cardinal direction substituted in for N/A/S/V

queries/bus.py
sultur: Good point! Added the todo for the time being.

sultur:
(The workflows currently fail due to old resource files in straeto package)

queries/grammars/bus.grammar
```diff
@@ -243,13 +243,8 @@ QBusWord/fall →
     | 'tólfa:kvk'_et_gr/fall
 
 QBusNumber/fall →
-    QBusNounSingular/fall QBusNr? QBusNumberWord
+    QBusNounSingular/fall QBusNr? UHeilTala
+    # UHeilTala is a nonterminal from the util modules, which
+    # matches any cardinal number and places it in the list result.numbers
```
vthorsteinsson, line 248: And does it allow "eitt", like the old QBusNumberWord did?

speech/__init__.py
```diff
@@ -166,8 +172,8 @@ def handle_endtag(self, tag: str):
                 dattrs = self._attr_stack.pop()  # Current tag attributes
                 t: Optional[str] = dattrs.pop("type")
                 assert t, f"Missing type attribute in <greynir> tag around string: {s}"
-                # Fetch handler
-                hf = self._handlers.get(t)
+                # Fetch handler function
+                hf = getattr(self._handler, t)
```
vthorsteinsson, line 176: Is hf typed somewhere?

speech/norm/__init__.py
```diff
@@ -77,7 +76,7 @@ def gssml(data: Any = None, *, type: str, **kwargs: Union[str, int, float]) -> s
     )
 
 
-# Spell out how character names are pronounced in Icelandic
+# Spell out how character names are pronunced in Icelandic
```
vthorsteinsson, line 79: pronounced is the correct spelling

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
-        "time": _time_handler,
-        "date": _date_handler,
-        "year": year_to_text,
-        "years": years_to_text,
-        "abbrev": _abbrev_handler,
-        "email": _email_handler,
-        "break": _break_handler,
-        "paragraph": lambda txt: f"<p>{txt}</p>",
-        "sentence": lambda txt: f"<s>{txt}</s>",
+class NormalizationHandler:  # Abstract base class
```
vthorsteinsson, line 185: It is possible to inherit from `abc.ABC` to create a proper abstract class that will raise a standard exception if an attempt is made to instantiate it

vthorsteinsson: ...and it would be better to actually define the abstract base interface on the base class, with the member functions raising `NotImplementedError`.

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
+    for Icelandic speech synthesis.
+    """
+
+    # TODO
+    # amount/s
+    # currency/ies
+    # distance/s
+
+    @classmethod
+    def number(cls, txt: str, **kwargs: str):
```
vthorsteinsson, line 201: Missing return types on all of these member functions

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
+        # Some times
+        if h == 0:
+            # Call 00:00 "tólf á miðnætti"
+            # and 00:xx "tólf ... um nótt"
+            h = 12
+            suffix = "á miðnætti" if m == 0 else "um nótt"
+        elif 0 < h <= 5:
+            suffix = "um nótt"
+        elif h == 12 and m == 0:
+            suffix = "um hádegi"
```
vthorsteinsson, line 260: "á hádegi" would be better and conform to "á miðnætti"

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
+            ),
+            None,  # Default if no matches are found
+        )
+        assert m is not None, f"Incorrect date format specified for date handler: {txt}"
+
+        # Handle 'DD/MM/YYYY' or 'MM. jan/feb/... [year]' match
+        gd = m.groupdict()
+        day = number_to_ordinal(gd["day"], gender="kk", case=case, number="et")
+        # Month names don't change in different declensions
+        month = (
```
vthorsteinsson, line 297: A little bit inefficient to have three lookups of gd["month"] here; a temp variable might be in order

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
+    @classmethod
+    def timespan(cls, seconds: str) -> str:
+        """Voicify a span of time, specified in seconds."""
+        # TODO: Replace time_period_desc in queries/util/__init__.py
+        raise NotImplementedError()
+
+    @classmethod
+    def year(cls, txt: str, *, after_christ: Optional[str] = None):
+        """Voicify a year."""
+        ac = after_christ is not None and after_christ == "True"
```
vthorsteinsson, line 317: A better term is "common era" instead of "after christ", cf. "BCE" = "Before Common Era".

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
     }
-)
+
+    @classmethod
+    def spell(cls, txt: str):
+        """Spell out a sequence of characters."""
+        if not txt:
+            return ""
+        t = [cls._CHAR_PRONUNCIATION.get(c.lower(), c) if c != " " else "" for c in txt]
+        s = ", ".join(t).strip()
```
vthorsteinsson, line 372: The strip() is probably unnecessary here

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
+        "z": "seta",
     }
-)
+
+    @classmethod
+    def spell(cls, txt: str):
+        """Spell out a sequence of characters."""
+        if not txt:
+            return ""
+        t = [cls._CHAR_PRONUNCIATION.get(c.lower(), c) if c != " " else "" for c in txt]
```
vthorsteinsson, line 371: The check for `c != " "` could be replaced by `!c.isspace()`, which would also consider tab, newline, CR, nonbreaking space, etc.

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
+        s = ", ".join(t).strip()
+        return re.sub(r"\s\s+", r" ", s)  # Shorten repeated whitespace
```
vthorsteinsson, line 373: The `re.sub()` is probably not needed here

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
+
+    @classmethod
+    def abbrev(cls, txt: str):
+        """Spell out a sequence of characters."""
+        return cls.spell(txt)
+
+    _DOMAIN_PRONUNCIATIONS = {
```
vthorsteinsson, line 380: type annotation: Mapping[str, str]

speech/norm/num.py
```diff
@@ -649,7 +650,7 @@ def number_to_ordinal(
 def numbers_to_ordinal(
     s: str,
     *,
-    regex: str = r"\b\d+\.(?=[ ,)])",
+    regex: str = r"((?<!\d\.)-)?\b\d+\.(?=[ ,)-])",
```
vthorsteinsson, line 653: I'm gonna trust you on this one ;-)

speech/voices/azure.py
```diff
@@ -212,3 +210,34 @@ def text_to_audio_url(
...
+
+class Normalization(DefaultNormalization):
+    """
+    Normalization handler class,
+    specific to the Azure voice engine.
+    """
+
+    # Override some character pronunciations for
+    # normalization, custom for this speech engine
+    _CHAR_PRONUNCIATION = {
```
vthorsteinsson, line 223: : Mapping[str, str]

speech/norm/__init__.py
```diff
@@ -231,42 +182,244 @@ def _email_handler(email: str) -> str:
...
+        return year_to_text(txt, after_christ=ac)
+
+    @classmethod
+    def years(cls, txt: str, *, after_christ: Optional[str] = None):
+        """Voicify text containing multiple years."""
+        ac = after_christ is not None and after_christ == "True"
+        return years_to_text(txt, after_christ=ac)
+
+    # Pronunciation of character names in Icelandic
+    _CHAR_PRONUNCIATION = {
```
vthorsteinsson, line 327: : Mapping[str, str]

speech/voices/azure.py
```diff
@@ -212,3 +210,34 @@ def text_to_audio_url(
...
+        "b": "bjé",
+        "c": "sjé",
+        "d": "djé",
+        "ð": "eeð",
+        "e": "eeh",
+        "é": "jé",
+        "g": "gjé",
+        "i": "ii",
+        "j": "íoð",
+        "o": "úa",
```
vthorsteinsson, line 234: Really?!

tests/test_queries.py
```diff
@@ -263,7 +264,128 @@ def test_bus(client: FlaskClient) -> None:
...
+    #     },
+    #     "WhichRoute",
+    # )
+    # assert "Naustabraut / Davíðshaga A" in json["answer"]
+    # assert (
+    #     all(not c.isdecimal() for c in json["voice"])
+    #     and "Naustabraut / Davíðshaga austur" in json["voice"]
+    # )
+
+    _query_data_cleanup()  # Remove any data logged to DB on account of tests
```
vthorsteinsson, line 388: I thought we had somewhere a test parameter that could be set on the query module which prevents logging?

tests/test_speech.py
```diff
@@ -571,20 +634,22 @@ def test_norm_digits() -> None:
 
 
 def test_norm_time_handler() -> None:
-    from speech.norm import _time_handler  # type: ignore
+    from speech.norm import DefaultNormalization
 
+    assert DefaultNormalization.time(f"00:00") == "tólf á miðnætti"
```
vthorsteinsson, line 639: Add "tólf á hádegi"

queries/bus.py
```diff
@@ -301,81 +260,153 @@ def QBusWord(node: Node, params: QueryStateDict, result: Result) -> None:
...
+
+    if split_symb is None:
+        # Bus stop is single noun phrase
+        np = NounPhrase(np).dative or np
+    else:
+        # Bus stop consists of two (at least) noun phrases
+        # separated by split_symb, inflect them separately
+        new_np: List[str] = []
+        for n in np.split(split_symb):
+            if any(c.islower() for c in n):
```
vthorsteinsson, line 370: Why not `if not n.isupper():`?

queries/bus.py
```diff
@@ -301,81 +260,153 @@ def QBusWord(node: Node, params: QueryStateDict, result: Result) -> None:
...
+
+    if split_symb is None:
+        # Bus stop is single noun phrase
+        np = NounPhrase(np).accusative or np
+    else:
+        # Bus stop consists of two (at least) noun phrases
+        # separated by split_symb, inflect them separately
+        new_np: List[str] = []
+        for n in np.split(split_symb):
+            if any(c.islower() for c in n):
```
vthorsteinsson, line 340: Why not `if not n.isupper():`?

queries/bus.py
```diff
@@ -589,65 +614,63 @@ def query_arrival_time(query: Query, session: Session, result: Result):
...
+                    va.append("kemur á")
+                va.append(accusative_form(stop.name, voice=True))
             va.append("klukkan")
             a.append("klukkan")
             if len(times) == 1 or (
                 len(times) > 1 and hms_diff(times[0], hms_now) >= 10
             ):
                 # Either we have only one arrival time, or the next arrival is
-                # at least 10 minutes away: only pronounce one time
+                # at least 10 minutes away: only pronunce one time
```
vthorsteinsson, line 644: Should be "pronounce" - that is the correct spelling

queries/bus.py
```diff
@@ -667,72 +690,58 @@ def query_arrival_time(query: Query, session: Session, result: Result):
         bq = bq.replace(*t)
     query.set_beautified_query(bq)
 
-    def assemble(x):
+    def assemble(x: Iterable[str]):
```
vthorsteinsson, line 693: Missing return type str

queries/bus.py
```diff
@@ -753,29 +762,16 @@ def sentence(state: QueryStateDict, result: Result) -> None:
...
-                    # We have both a normal and a voice answer
-                    response, answer, voice_answer = response
-                assert answer is not None
-                q.set_answer(response, answer, voice_answer)
-            except AssertionError:
+        try:
+            assert qfunc is not None, "qfunc is None"
+            q.set_answer(*qfunc(q, result))
+        except AssertionError:
+            raise
```
vthorsteinsson, line 771: Is this definitely what we want in production?

queries/grammars/bus.grammar
sultur: Yep!
Should I add a specifier though so it only matches numbers in `_nf_hk`?
It would then only match queries like `... leið (númer) eitt` or `... leið (númer) tvö` but not `... leið (númer) einn` or `leið (númer) þrír`.

speech/norm/__init__.py
sultur: This argument adds `"eftir Krist"` to the end of the year (might be used in certain situations dealing with historical text, not common otherwise).
I thought that this name would better indicate the effect, otherwise people might enable the argument in every case where the year is in the common era.

speech/norm/__init__.py
sultur: Fixed.

speech/norm/__init__.py
sultur: Fixed!

speech/norm/__init__.py
sultur: Fixed

speech/norm/__init__.py
sultur: fixed

speech/norm/__init__.py
sultur: Good point, fixed.

speech/norm/__init__.py
sultur: True, fixed.

speech/norm/__init__.py
sultur: Added!

speech/voices/azure.py
sultur: Yep :/ the Azure speech synthesis pronounces stuff weird, I found these to be closest to the actual pronunciation through trial and error.

tests/test_speech.py
sultur: Added

queries/bus.py
sultur: Good point, fixed!

queries/bus.py
sultur: added

queries/bus.py
sultur: Probably not, no. I didn't modify this in the pull request though. I guess I'll just remove the `except AssertionError: ...` thing for now.

tests/test_queries.py
sultur: In my experience it can be a bit flaky sometimes. When I interrupt a running test, sometimes stuff ends up in the database. This at least explicitly cleans up db stuff when the bus test function finishes.
I will look into this issue soon though (and verify whether or not I'm just imagining things being added to the db :) ).

queries/grammars/bus.grammar
sultur: Decided to allow incorrect cases/genders for the bus numbers (the queries are understandable and nonambiguous).

speech/__init__.py
sultur: It's a bit tricky to type hint this, I added a simple `(...) -> str` type for now.
I couldn't find a way to correctly specify "_classmethod which receives strings as keyword arguments_", but I'll fix this in the next pull request (which I'm currently working on).

speech/norm/__init__.py
sultur: I decided to remove the `NormalizationHandler` base class and instead just use `DefaultNormalization` as the base class.
I didn't see a need for defining an abstract base class (with some kind of function interface) at the moment, but maybe that will prove useful when the normalization functionality is developed further.

PR #202, opened 2022-12-19
Added ordinal grammar support

hrolfurinn:
Added parsing of ordinal numbers written out (i.e. spoken by user) in queries/util. The value of the spoken ordinal is collected, similar to handling of other numbers previously supported. Also fixed some grammar issues with the support of other numbers.


sveinbjornt (commented):
LGTM

PR #203, opened 2022-12-21
Query refactoring

sveinbjornt:
* A pedestrian refactoring and review PR
* Migrated query module code to f-strings
* Improved query module tests
* Assorted little fixes, rm unused code etc.
* Not really much in the way of changes to logic


tests/test_queries.py
```diff
@@ -880,18 +882,18 @@ def test_opinion(client: FlaskClient) -> None:
     assert json["key"] == "þriðji orkupakkinn"
 
     json = qmcall(
-        client, {"q": "hvað finnst þér eiginlega um Katrínu Jakobsdóttur"}, "Opinion"
+        client, {"q": "hvað finnst þér eiginlega um Sigmund Davíð"}, "Opinion"
     )
     assert json["answer"].startswith("Ég hef enga sérstaka skoðun")
-    assert json["key"] == "Katrín Jakobsdóttir"
+    assert json["key"] == "Sigmundur Davíð"
```
sultur, line 888: :joy_cat:

tests/test_queries.py
```diff
@@ -910,13 +912,14 @@ def test_petrol(client: FlaskClient) -> None:
...
 def test_pic(client: FlaskClient) -> None:
-    """Pic module"""
+    """Pic module."""
 
     if not has_google_api_key():
         # NB: No Google API key on test server
         return
 
-    json = qmcall(client, {"q": "Sýndu mér mynd af Katrínu Jakobsdóttur"}, "Picture")
+    # TODO: Re-add test with "Katrín Jakobsdóttir" when fixed GreynirPackage is released
```
sultur, line 921: Do we want to handle this ambiguity in GreynirPackage though? I think we should somehow try to infer the correct root of the name by popularity in the Greynir scraper database.

My idea, roughly, would be to create a function which lists all possible names (in nf.) matching the provided form, search our database for matching mentions of the name, and return the most popular one.
The function could then be used in this case (when showing an image of a person) and also in builtin.py (when asking about someone), or even in the unnecessary key from the opinion module.
(This would have a bit of overhead though, maybe we can optimize this in some way.)

sultur (approved):
lgtm!

(There are still some `"<break ... />"` strings in the query modules (counting, flights, wiki, words) which are changed to f-strings in this PR, but they get removed in my speech normalization PR.)

PR #208, opened 2023-03-15
Changes to gitignore. Fix psql setup instructions for versions >=15

sultur:
- Changes to gitignore (added `.env` and everything in `resources/` except `resources/geo/`).
- Added instructions to fix psql 15+ permission issue in setup docs.
- Added python-dotenv as dependency which allows setting env variables in the file `.env`.


sultur:
Merging, as I want to re-deploy and these are very minor changes.

PR #211, opened 2023-04-18
Added query authentication status property

sveinbjornt:
Query modules can now check if a query request is authenticated with a Greynir Server Key, e.g.

````
q.authenticated
````


sveinbjornt:
@vthorsteinsson

PR #216, opened 2023-08-21
Íslandsbanki atm query module

KariSt1:
Query functionality to locate nearest Íslandsbanki atm and get further information about its services.


queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+    if not atm_data:
+        return None
+
+    filtered_atm_data: List[Dict[str, Any]] = []
+    for s in atm_data:
+        """Filter out all non ATMs from the dictionary"""
+        item_type: str = s.get("type", "")
+        if item_type == "atm":
+            filtered_atm_data.append(s)
+    atm_data = filtered_atm_data
```
sultur, line 175: This (the filtering) can be included in `_get_atm_data()`. Then we can also cache the results so we don't read the file from disk each time a query is performed. I'd make `_get_atm_data` return the `filtered_atm_data` variable, and then add the following decorator:
```py
import cachetools

...

_TTL = 86400 # 1 day

@cachetools.cached(cachetools.TTLCache(1, _TTL))
def _get_atm_data():
    ...
```
This caches the ATM locations in memory for one day (86400 seconds).
We still need to do the distance calculations each time a query is made though.

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+    filtered_atms: List[Dict[str, Any]] = []
+    for atm in atms:
+        services = atm.get("services", {})
+        if services.get("coinmachine", False) is True:
+            filtered_atms.append(atm)
+    atms = filtered_atms
+
+    # Sort by distance
+    dist_sorted = sorted(atms, key=lambda s: s["distance"])
+    return _group_closest_based_on_address(dist_sorted)
```
sultur, line 267: These three lines are repeated in lines 211-213 & 247-249, couldn't we just include the `sorted(atms, ...)` call as the first step in `_group_closest_based_on_address()`?

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+            # replace irregular hyphen with regular hyphen
+            last_word.replace("–", "-")
+        if "-" in last_word:
+            split_nr = last_word.split("-")
+            first_number = number_to_text(split_nr[0], case="þf")
+            last_number = number_to_text(split_nr[-1], case="þf")
+            number = first_number + " til " + last_number
+            return " ".join(word_list[:-1]) + " " + number
+        return " ".join(word_list[:-1]) + " " + number_to_text(last_word, case="þf")
+    return s
```
sultur, line 290: Ah, I think I forgot to mention the code we have for transcribing stuff for TTS in `speech/trans`, like the `gssml` function (you can see usage of it in e.g. the arithmetic module), but I notice now that we haven't implemented a method for transcribing addresses yet. This looks good for the time being though :).

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+_ERRMSG = "Ekki tókst að sækja upplýsingar um hraðbanka."
+
+
+def _answ_for_atm_query(q: Query, result: Result) -> AnswerTuple:
+    """Return an answer tuple for the given ATM query"""
+    ans_start = ""
+    atm_list: List[Dict[str, Any]] = []
+    if "context_reference" in result and result.last_atm is not None:
+        # There is a reference to a previous result
+        # TODO: Handle multiple ATMs in same location, but with different services
```
sultur, line 324: Could you move the TODO comments to the top of the file, between the module docstring and the imports? Handy for quickly assessing what needs to be done in each query module (usually a lot of things).

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+        if result.qkey == "ClosestAtm":
+            atm_list = _closest_atm(location)
+            ans_start = "Næsti hraðbanki"
+        elif result.qkey == "ClosestAtmDeposit":
+            atm_list = _closest_atm_deposit(location)
+            ans_start = "Næsti hraðbanki sem tekur við innborgunum"
+        elif result.qkey == "ClosestAtmForeignExchange":
+            atm_list = _closest_atm_foreign_exchange(location)
+        elif result.qkey == "ClosestAtmCoinmachine":
+            atm_list = _closest_atm_coinmachine(location)
```
sultur, line 354: I cannot wait for the `match-case` statement introduced in Python 3.10!
Nice job keeping the `if-elif` statements compact and understandable :)
(They can become quite unwieldy in the query modules due to icelandic declensions)

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+            answ_fmt: str = (
+                "Ekki er hægt að kaupa erlendan gjaldeyri í hraðbankanum við {0}."
+            )
+            answer: str = answ_fmt.format(
+                street_name,
+            )
+            voice: str = answ_fmt.format(
+                voice_street_name,
+            )
+    elif result.qkey == "AtmFurtherInfoCoinmachine":
```
sultur, line 521: _Very minor point:_ It may be a good idea to make the qkeys referenced in this function constants, and reference those (also in the corresponding nonterminal functions). Just decreases the likelihood of minor typos suddenly eradicating functionality from a query module (those can be a joy to debug :smiling_face_with_tear:).

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+                )
+            else:
+                ans_start = f"Nei, ekki er hægt að leggja inn á reikninginn þinn í {atm_word} við "
+        elif result.qkey == "AtmFurtherInfoWithdrawalLimit":
+            ans_start = f"Hámarksúttekt í {atm_word} við "
+
+    elif "error_context_reference" in result:
+        return gen_answer("Ég veit ekki til hvaða hraðbanka þú vísar.")
+    else:
+        location = q.location
```
sultur, line 341: Couldn't we just pass in the location as an argument instead of `q: Query`? We already check `q.location` in the `sentence(...)` function.

sultur (changes requested):
Great work on this module!

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+    natlang_seq,
+    read_grammar_file,
+    AnswerTuple,
+    LatLonTuple,
+)
+from speech.trans.num import number_to_text
+
+_ATM_QTYPE = "Atm"
+
+_PATH_TO_ISB_JSON = "../resources/geo/isb_locations.json"
```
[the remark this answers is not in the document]
sveinbjornt: Absolutely use Pathlib. I'm in the process of moving most of the geo.py code (which uses files in `resources/geo`) into the `iceaddr` external package so this directory might not be around for much longer. I suggest creating a `queries/resources` directory for resources used exclusively by query modules and not elsewhere in the codebase.

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+    "GBP": "sterlingspund",
+    "EUR": "evrur",
+    "DKK": "danskar krónur",
+    "SEK": "sænskar krónur",
+    "PLN": "pólsk slot",
+    "ISK": "íslenskar krónur",
+}
+
+
+def help_text(lemma: str) -> str:
```
sveinbjornt, line 63: This is function is not used unless you define TOPIC_LEMMAS like in some of the other modules. Maybe best to leave it out for now.

queries/atm.py
sveinbjornt: Great minds think alike.

sveinbjornt:
This looks good to me, but you should write some tests for this in `tests/test_queries.py`

queries/atm.py
sveinbjornt: Or add in the TOPIC_LEMMA "hraðbanki" and more examples of queries supported by this module in random help_text

queries/atm.py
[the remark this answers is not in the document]
KariSt1: Fixed.

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+
+
+def _read_isb_location_data_from_file() -> List[Dict[str, Any]]:
+    """Read isb locations JSON data from file"""
+    try:
+        script_dir = os.path.dirname(__file__)
+        rel_path = _PATH_TO_ISB_JSON
+        abs_file_path = os.path.join(script_dir, rel_path)
+        with open(abs_file_path, "r", encoding="utf-8") as f:
+            return json.load(f)
```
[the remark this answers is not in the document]
KariSt1: Fixed, now using pathlib.

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+    except Exception as e:
+        logging.error("Error reading isb locations json data: %s", e)
+        return []
+
+
+def _get_atm_data() -> Optional[List[Dict[str, Any]]]:
+    """Get ATM data"""
+    atm_data: List[Dict[str, Any]] = _read_isb_location_data_from_file()
+    return atm_data
```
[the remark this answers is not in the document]
KariSt1: Fixed. Added caching to this function and added filtering to it so that it only keeps the atms from the file.

queries/atm.py
KariSt1: Fixed.

queries/atm.py
KariSt1: Fixed. Added constants for the qkeys.

queries/atm.py
KariSt1: Fixed, now passing the location as an argument instead of the query.

queries/atm.py
KariSt1: Fixed. Added a topic lemma and further query examples.

queries/atm.py
```diff
@@ -0,0 +1,571 @@
+"""
+
+    Greynir: Natural language processing for Icelandic
+
+    Atm query response module
```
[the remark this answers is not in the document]
KariSt1: Fixed.

queries/atm.py
KariSt1: Fixed, the sort call is now the first step in `_group_closest_based_on_address()`.

queries/atm.py
```diff
@@ -0,0 +1,571 @@
...
+        """Filter out all non ATMs from the dictionary"""
+        item_type: str = s.get("type", "")
+        if item_type == "atm":
+            filtered_atm_data.append(s)
+    atm_data = filtered_atm_data
+    if loc:
+        # Calculate distance of all stations
+        for s in atm_data:
+            s["distance"] = distance(
+                loc, (s["location"]["latitude"], s["location"]["longitude"])
```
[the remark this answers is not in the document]
KariSt1: Fixed. Longitude, latitude, and street address checks are now made when filtering out the atms in the `get_atm_data` function so there should never be a `KeyError` here anymore.

sultur (approved):
LGTM, once the tests pass and @sveinbjornt approves we can merge this.

PR #217, opened 2023-08-30
Updating tests to use skipif

KariSt1:
Tests now skip instead of passing if an api key or a file is required for the test but not present. Also renamed `read_api_key` to `read_txt_api_key` and added a `read_json_api_key` function to `utility.py` as not all keys are .txt files. These methods now allow for an optional `folder` parameter to specify if the key is not in the default `resources` folder. Tests for these methods were added to `test_util.py`.


sveinbjornt:
Good work!

